PR #1, opened 2026-01-28
Update rust SDK to use new Hegel protocol

DRMacIver:
This is the SDK-side change from https://github.com/antithesishq/hegel/pull/1

This is even more Claude-written than the Python one and requires correspondingly more care. I got the Python SDK to the point where I was happy with it, and then asked Claude to go wild on the Rust one. Will similarly prereview.

### DIFF
--- a/.claude/reliability-config.yaml
+++ b/.claude/reliability-config.yaml
@@ -1,3 +1,0 @@
-git_repo: true
-require_push: true
-auto_work_on_tasks: true
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ target/
 **/.claude/
 result
 .claude-reliability/
+.claude/reliability-config.yml
 
 # claude-reliability managed
 .claude/bin/


.claude/reliability-config.yaml
```diff
@@ -0,0 +1,3 @@
+git_repo: true
```
Liam-DeVoe: gitignore this file

src/runner.rs
```diff
@@ -0,0 +1,620 @@
...
+            Verbosity::Quiet => "quiet",
+            Verbosity::Normal => "normal",
+            Verbosity::Verbose => "verbose",
+            Verbosity::Debug => "debug",
+        }
+    }
+}
+
+/// Special marker used to identify assume(false) panics.
+const REJECT_MARKER: &str = "HEGEL_REJECT";
```
Liam-DeVoe, line 180: I haven't looked into this, but why is this a panic? I'd expect communicating Status.INVALID test cases should be part of the cbor protocol?

src/runner.rs
```diff
@@ -0,0 +1,620 @@
...
+) {
+    // Set thread-local state for this test case
+    // Note: we pass the channel directly (not cloned) so generators and mark_complete
+    // share the same message ID sequence.
+    set_is_last_run(is_final);
+    set_connection(Arc::clone(connection), test_channel);
+
+    // Run test in catch_unwind
+    let result = catch_unwind(AssertUnwindSafe(test_fn));
+
```
Liam-DeVoe, line 518: `catch_unwind` seems bad here, it doesn't seem like the protocol should ever throw, valid/interesting status should be communicated purely via cbor messages

src/runner.rs
DRMacIver: Yes and then you panic to abort the test.

This is somewhat unidiomatic for rust, but I think we should just eat the idiomaticity penalty here. We have to be set up to handle panics anyway, and users are not supposed to recover from aborted tests, so having this sort of thing be a Result instead of a panic is essentially fake idiomatic even if we do it because you're not allowed to do any of the things Result implies.

src/runner.rs
DRMacIver: Rust tests fail by assert!, so an idiomatic rust test failure _is_ a panic, which we then catch and use to mark interesting. We're client side so we're deciding what's interesting, not the server.

Liam-DeVoe: great point. we're catching the rust-side exception for control flow, not the protocol exception. I had a wrong mental model.

src/runner.rs
Liam-DeVoe: yup agree we just have to catch for control flow. disregard my comment

PR #18, opened 2026-02-24
Add support for seeds

echoumcp1:
Title.

Winter fixed the flake.nix to remove non-idiomatic flake usage for Rust projects.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
 RELEASE_TYPE: patch
 
-This patch adds support for setting `seed` to the protocol.
+This patch adds support for setting `seed` as an option to `hegel`.


src/runner.rs
```diff
@@ -222,6 +222,7 @@ pub struct Hegel<F> {
     test_fn: F,
     test_cases: u64,
     verbosity: Verbosity,
+    seed: Value,
```
Liam-DeVoe: Is this a reference to serde's `Value`? This should be an Option<u64> instead

src/runner.rs
```diff
@@ -249,6 +251,11 @@ where
         self
     }
 
+    pub fn seed(mut self, seed: i64) -> Self {
+        self.seed = Value::Integer(seed.into());
+        self
+    }
+
```
Liam-DeVoe, line 258: this should support setting `seed` to `None`

src/runner.rs
```diff
@@ -372,6 +379,7 @@ where
             "command" => "run_test",
             "name" => "test",
             "test_cases" => self.test_cases,
+            "seed" => self.seed.map_or(Value::Null, |seed| Value::Integer(seed.into())),
```
Liam-DeVoe: Claude suggests the cleaner `"seed" => self.seed.map_or(Value::Null, Value::from),` here?

src/runner.rs
```diff
@@ -249,6 +251,11 @@ where
         self
     }
 
+    pub fn seed(mut self, seed: Option<i64>) -> Self {
```
Liam-DeVoe: seeds are consistently `u64` in the rust ecosystem, we should match that here

PR #20, opened 2026-02-25
Add getting-started docs and CI rationalization

DRMacIver:
## Summary

- Adds `docs/getting-started.md` following the standard 12-section structure
- Splits monolithic `ci.yml` into per-check workflow files (check, coverage, conformance, nix, check-release, release, zizmor)
- Replaces `extractions/setup-just@v2` with curl-based just installation
- Pins all GitHub Actions to SHA hashes and fixes zizmor security findings (persist-credentials, template injection, permissions)
- Adds `zizmor.yml` workflow for GitHub Actions linting
- Updates justfile with standard recipes: `setup`, `build-conformance`, `conformance`, `test-all-features`
- Adds RELEASE.md for the changes
- Slims down README.md to point to getting-started docs

## Test plan

- [ ] `just check` passes (lint + docs + test + test-all-features)
- [ ] `just conformance` passes
- [ ] All workflow files are valid YAML
- [ ] `zizmor --format plain .github/` reports no findings (except expected `dangerous-triggers` on release)

### DIFF
--- a/justfile
+++ b/justfile
@@ -51,5 +51,6 @@ build-conformance:
     cargo build --release --manifest-path tests/conformance/rust/Cargo.toml
 
 conformance: build-conformance
-    uv run --with pytest --with pytest-subtests --with hypothesis \
+    uv run --with "hegel @ git+ssh://git@github.com/antithesishq/hegel.git" \
+        --with pytest --with pytest-subtests --with hypothesis \
         pytest tests/conformance/test_conformance.py --durations=20 --durations-min=1.0


.github/workflows/ci.yml
```diff
@@ -58,72 +39,17 @@ jobs:
...
 
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
+          components: rustfmt, clippy
       - uses: Swatinem/rust-cache@v2
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
```
Liam-DeVoe, line 112: we should chat about how to structure justfiles and ci worfklows. I had intentionally split out eg docs as its own job, both for CI speed and because it's nice to see exactly which job failed.

.github/workflows/ci.yml
```diff
@@ -135,28 +61,20 @@ jobs:
       - uses: ./.github/actions/base
         with:
           ssh-key: ${{ secrets.HEGEL_DEPLOY_KEY }}
+
+      - uses: extractions/setup-just@v2
```
Liam-DeVoe: I'm not a fan of relying on a 100-star github action (supply chain risk, etc). I prefer the one-line `sudo apt-get update && sudo apt-get install -y just`.

.github/workflows/ci.yml
DRMacIver: Yes, agreed these should be separate jobs. I asked Claude to rationalise the CI jobs and it very much did the opposite of the correct thing and consolidated them all.

.github/workflows/ci.yml
DRMacIver: I have a vaguely cached opinion that apt-get just isn't good, but I'm not sure where that's from. Agreed we shouldn't be using this action.

justfile
```diff
@@ -13,27 +39,18 @@ check-format-nix:
...
+
+build-conformance:
+    cargo build --release --manifest-path tests/conformance/rust/Cargo.toml
 
-# aliases for local developer experience. CI and builds should use the longer names.
-alias test := check-test
-check: check-format lint check-test check-docs
+conformance: build-conformance
+    uv run --with "hegel @ git+ssh://git@github.com/antithesishq/hegel.git" \
+        --with pytest --with pytest-subtests --with hypothesis \
```
Liam-DeVoe, line 55: `pytest-subtests` shouldn't be required!

DRMacIver: Ugh. Sorry. I don't know why that keeps sneaking in.

PR #37, opened 2026-03-09
Manage hegel binary via pinned .hegel/venv

DRMacIver:
## Summary

- `HEGEL_VERSION` is now a plain git commit hash for hegel-core, hard-coded in `runner.rs`
- On first use, creates a project-local `.hegel/venv` virtualenv via `uv venv` and installs hegel into it via `uv pip install`
- A `.hegel/venv/hegel-version` file tracks the installed hash; subsequent runs skip installation if it matches `HEGEL_VERSION`
- On version mismatch the venv is recreated and hegel is reinstalled
- `HEGEL_CMD` environment variable overrides to a specific binary path, skipping installation entirely
- The release script pins `HEGEL_VERSION` to the current hegel-core main HEAD SHA at release time via `gh api`
- Repo URL updated from `hegel.git` to `hegel-core.git`

## Test plan

- [ ] CI passes (cargo test, clippy, conformance)
- [ ] Manual test: delete .hegel/, run cargo test, verify auto-install

### DIFF
--- a/tests/test_output.rs
+++ b/tests/test_output.rs
@@ -26,7 +26,7 @@ fn test_failing_test_output() {
     //   Draw 1: 0
     //   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
     let expected = Regex::new(concat!(
-        r"^thread '.*' \(\d+\) panicked at src/main\.rs:\d+:\d+:\n",
+        r"thread '.*' \(\d+\) panicked at src/main\.rs:\d+:\d+:\n",
         r"intentional failure: -?\d+\n",
         r"Draw 1: -?\d+",
     ))
@@ -62,7 +62,7 @@ fn test_failing_test_output_with_backtrace() {
     //   note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
     let expected = Regex::new(concat!(
         r"(?s)",
-        r"^thread 'main' \(\d+\) panicked at src/main\.rs:\d+:\d+:\n",
+        r"thread 'main' \(\d+\) panicked at src/main\.rs:\d+:\d+:\n",
         r"intentional failure: -?\d+\n",
         r"Draw 1: -?\d+\n",
         r"stack backtrace:\n",
@@ -96,7 +96,7 @@ fn test_failing_test_output_with_full_backtrace() {
 
     let expected = Regex::new(concat!(
         r"(?s)",
-        r"^thread 'main' \(\d+\) panicked at src/main\.rs:\d+:\d+:\n",
+        r"thread 'main' \(\d+\) panicked at src/main\.rs:\d+:\d+:\n",
         r"intentional failure: -?\d+\n",
         r"Draw 1: -?\d+\n",
         r"stack backtrace:\n",


.github/workflows/ci.yml
```diff
@@ -72,22 +72,25 @@ jobs:
...
         with:
           python-version: "3.13"
       - name: Install hegel
-        run: pip install "git+ssh://git@github.com/antithesishq/hegel.git"
+        run: pip install "git+ssh://git@github.com/antithesishq/hegel-core.git"
 
       - name: Run tests
         run: RUST_BACKTRACE=1 cargo test
+        env:
+          HEGEL_CMD: hegel
```
Liam-DeVoe: isn't the right way to just install `uv` and let hegel-rust install from hegel-core automatically? I don't see why the CI jobs need to set eg `HEGEL_CMD` or manually install this

.github/scripts/release.py
```diff
@@ -100,6 +100,24 @@ def check(base_ref: str) -> None:
     parse_release_file(release_file)
 
 
+def pin_hegel_version(runner_rs: Path) -> None:
+    """Pin HEGEL_VERSION to the current HEAD of hegel-core main."""
+    sha = subprocess.check_output(
+        ["gh", "api", "repos/antithesishq/hegel-core/commits/main", "--jq", ".sha"],
+        text=True,
```
Liam-DeVoe, line 107: It's a bad idea to pin to HEAD, which will result in subtle inconsistencies compared to releases. We should instead pin to the latest release. 

This also means `HEGEL_VERSION` should be a git tag name, not a commit ref, so humans can actually inspect the diff and make sure it's correct. If we let it be a commit hash nobody is ever going to double check it and we're probably going to end up with a bad mismatch after a tooling bug at some point.

Liam-DeVoe: once we're on pypi we should use that as the source of truth instead of git tags (TODO to self: make an issue for this if we agree on this)

DRMacIver: Yup, agreed on this.

.github/workflows/ci.yml
DRMacIver: Yeah, sorry, Claude being weird and my usual problem where my eyes glaze over every time I look at a github action yml file. You're clearly correct here.

.github/workflows/ci.yml
```diff
@@ -43,17 +43,16 @@ jobs:
...
       - uses: Swatinem/rust-cache@ad397744b0d591a723ab90405b7247fac0e6b8db # v2
 
       - name: Check formatting
         run: cargo fmt --check
       - name: Clippy
         run: cargo clippy --all-features --tests -- -D warnings
       - name: Install zizmor
-        run: pip install zizmor
+        run: pipx install zizmor==1.23.1
       - name: Run zizmor
```
Liam-DeVoe, line 56: I have a personal dislike of `pipx` and prefer just installing to the global python env, which I recognize is technically polluting the env for proceeding calls but in practice I've never had a problem, and results in one less tool to learn. I'm not pushing to change this, just registering my belief. Fine to stick with pipx 👍

.github/workflows/ci.yml
```diff
@@ -74,17 +73,11 @@ jobs:
         with:
           ssh-key: ${{ secrets.HEGEL_DEPLOY_KEY }}
 
-      - uses: dtolnay/rust-toolchain@efa25f7f19611383d5b0ccf2d1c8914531636bf9 # master
-        with:
-          toolchain: ${{ matrix.rust }}
```
Liam-DeVoe, line 79: hmm, I initially used `dtolnay/rust-toolchain` as it was recommended for caching reasons. The "install rust" takes 10 seconds on this PR though, so I guess it's just fine?

.github/workflows/ci.yml
```diff
@@ -222,6 +206,7 @@ jobs:
     if: github.event_name == 'push' && github.repository == 'antithesishq/hegel-rust'
     needs: [lint, test, test-all-features, docs, coverage, conformance, nix]
     runs-on: ubuntu-latest
+    environment: release
     permissions:
       contents: write
```
Liam-DeVoe, line 210: remove this `environment`

src/runner.rs
```diff
@@ -15,8 +15,11 @@ use std::time::Duration;
 use tempfile::TempDir;
 
 const SUPPORTED_PROTOCOL_VERSIONS: (f64, f64) = (0.1, 0.3);
```
Liam-DeVoe, line 17: we can get rid of `SUPPORTED_PROTOCOL_VERSIONS` now. And replace it with...nothing, probably? I would normally say an assert that the returned version is what we expect, but we're pinning to a hegel-core version tag which is different than the protocol version. And I don't particularly see a benefit in maintaining a mapping between the two just for a client-side assertion we don't ever expect to hit.

Oh, but people can override the hegel path. Hmm so we do need this. We should add a CI job that tests with the lowest version of the hegel server that we claim to support. Which will require a mapping between these two somewhere...maybe hegel-core maintains that? Not sure.

src/runner.rs
```diff
@@ -148,6 +151,81 @@ fn init_panic_hook() {
     });
 }
 
+fn ensure_hegel_installed() -> Result<String, String> {
+    let venv_dir = format!("{HEGEL_SERVER_DIR}/venv");
+    let version_file = format!("{venv_dir}/hegel-version");
```
Liam-DeVoe, line 157: I would put `hegel-version` outside of the venv? it's a thing managed by us, not the venv, and is human-inspectable. `.hegel/hegel-version` seems the right place, not `.hegel/venv/hegel-version`.

.github/workflows/ci.yml
DRMacIver: I don't have a strong opinion either way. Mild preference for pipx but I've not used it more than very casually. Can you say more about why you dislike it?

.github/workflows/ci.yml
DRMacIver: zizmor apparently dislikes it (thinks it's redundant?). I'm happy to put it back in if you want, just needs updating the zizmor conf.

.github/workflows/ci.yml
DRMacIver: Claude thought that this was needed for the release action to work. I don't understand enough about github actions to be sure if it's right or wrong, but it seemed plausible. I guess the fact that it wasn't there before means that it's wrong though.

src/runner.rs
DRMacIver: Yeah, I wondered about this too. I'm not sure what the right thing to do here is and figured we'd leave it for later work.

I think another thing to consider here is that this is relevant to when we run on antithesis later.

src/runner.rs
DRMacIver: My reasoning on putting it inside the venv is that this means that you can delete the venv if you want to reset the hegel install (which people will do whether we like it or not) and this means that works correctly.

.github/workflows/ci.yml
Liam-DeVoe: it's just wrong yeah

.github/workflows/ci.yml
Liam-DeVoe: I think zizmor is right? not 100% sure. I definitely don't mind removing a github action if our runners already have the rust toolchain. (only concern is version pinning but I think rust sufficiently isolates rustup and rust versions that this is not a problem.)

Let's stick with this PR version

.github/workflows/ci.yml
Liam-DeVoe: Mostly fear of unknown unknowns - I have a basically perfect mental model of how pip works and no mental model of how pipx works. Arguably this is a skill issue on my end.

src/runner.rs
Liam-DeVoe: happy to defer this, I'll open an issue

src/runner.rs
Liam-DeVoe: hmm. I think people are likely to delete the entire `.hegel` than just the venv, but it's true some will delete just the venv. 

I don't think there's any fundamental reason we can't make this work though, if you have `.hegel/hegel-version` but no `.hegel/venv` dir we just recreate `.hegel/venv` with that version. The class of bugs tying them together fixes is where we do a bad job syncing updates to them. Which is definitely a possible bug we might write.

So I am ambivalent on this

DRMacIver: I think people might not want to delete .hegel because it has the database in it.

More generally: I think there's no really big downside to the version being in the venv and it makes a bunch of small classes of problem harder to happen.

.github/workflows/ci.yml
DRMacIver: I think my mental model of it is just that it creates a per-package virtualenv, installs into that virtualenv with pip, and puts the entry points from that package in that virtualenv on your path. That's what the project description says it does and I've never noticed any discrepancies from it, but tbf I've also not looked carefully enough that I'm confident I would have.

src/runner.rs
Liam-DeVoe: only downside is it's harder for a human poking around to find out what version they're pinned to

DRMacIver: I think only harder in the sense that they need to know it's there, though I do agree it's less obvious.

PR #44, opened 2026-03-11
Automatically add `database_key` in `#[hegel::test]`

Liam-DeVoe:
- [x] Depends on https://github.com/hegeldev/hegel-core/pull/59.

We use something similar to the fully-qualified test name in pytest by mixing in the module path and function name; e.g. `my_tests::test_addition_commutes`.

### DIFF
--- a/tests/test_antithesis.rs
+++ b/tests/test_antithesis.rs
@@ -17,7 +17,7 @@ fn my_test(tc: hegel::TestCase) {
 }
 "#;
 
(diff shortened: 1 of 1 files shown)


DRMacIver (approved):
I haven't read this as carefully as I'd like to under normal circumstances, but I'm keen to get this in, what I've skimmed looks sane enough, and I trust your judgement on the details.

PR #45, opened 2026-03-11
Fix hegel-core version and improve debuggability

DRMacIver:
The main purpose of this PR was to fix the library to work at all - we should no longer auto-update the dependency as part of the release process, which breaks things as we run that after the tests (in retrospect, duh) - but the experience of trying to debug that was so cursed because of our panic hook that this ended up going down a bit of a rabbit hole of improving the ergonomics of debugging hegel-rust problems in general.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -138,14 +138,11 @@ fn init_panic_hook() {
         let prev_hook = panic::take_hook();
         panic::set_hook(Box::new(move |info| {
             if !currently_in_test_context() {
-                // Outside a test case (e.g. setup, version negotiation, final panic) —
-                // use the previous hook so the message is visible.
+                // use actual panic hook outside of tests
                 prev_hook(info);
                 return;
             }
 
-            // Inside a test case — suppress output (we're likely shrinking)
-            // and store the info for later use in run_test_case.
             let thread = std::thread::current();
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -35,6 +35,15 @@ coverage:
...
+    #!/usr/bin/env bash
+    set -euo pipefail
+    tag=$(gh api repos/antithesishq/hegel-core/releases/latest --jq '.tag_name')
+    sed -i '' "s/^const HEGEL_SERVER_VERSION: &str = \".*\"/const HEGEL_SERVER_VERSION: \&str = \"${tag}\"/" src/runner.rs
+    echo "Updated HEGEL_SERVER_VERSION to ${tag}"
+    # Clear cached install so the next test run picks up the new version
+    rm -rf .hegel/venv
+
 build-conformance:
     cargo build --release --manifest-path tests/conformance/rust/Cargo.toml
```
Liam-DeVoe, line 49: why move this to bash instead of keep in python? Also nothing calls this at the moment that I can see

DRMacIver: No particular reason except it was being taken out of the release flow (which is in Python) and moved into the justfile (which is in ~bash).

And yes it's something that needs manually calling. We could have some sort of daily update PR flow to keep it up to date if we wanted but I haven't thought through the details.

PR #50, opened 2026-03-12
API Rationalization

DRMacIver:
This implements (I think) all of the big API changes we want to make prior to release, both ones we discussed and also some things that made me go "ugh that's gross" when looking at usage examples earlier.

Duplicating the RELEASE.md, key changes are:

* Move most things from top-level functions to methods on a `TestCase` object explicitly passed in
* `draw` now silently drops printing when not at the top level, and compose! just takes a TestCase object that it can call draw on.
* New `draw_silent` method that doesn't ever print, and lacks the Debug bound.
* Replace use of from_type<T>() with T::default_generator(). The from_type did nothing useful and often broke type inference.
* Rename Generate trait to Generator
* Make draw take generators by value, so you no longer need draw(& most of the time (A reference to a Generator is a Generator though, so you can still pass by reference if you want)

### DIFF
--- a/src/generators/collections.rs
+++ b/src/generators/collections.rs
@@ -436,7 +436,9 @@ pub fn arrays<G: Generator<T> + Send + Sync, T, const N: usize>(
     ArrayGenerator::new(element)
 }
 
-impl<G: Generator<T> + Send + Sync, T, const N: usize> Generator<[T; N]> for ArrayGenerator<G, T, N> {
+impl<G: Generator<T> + Send + Sync, T, const N: usize> Generator<[T; N]>
+    for ArrayGenerator<G, T, N>
+{
     fn do_draw(&self, tc: &TestCase) -> [T; N] {
         if let Some(basic) = self.as_basic() {
             basic.do_draw(tc)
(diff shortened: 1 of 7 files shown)


src/generators/generators.rs
```diff
@@ -1,11 +1,11 @@
...
+use crate::test_case::{labels, TestCase};
 use ciborium::Value;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
 /// A bundled schema + parse function for schema-based generation.
 ///
-/// The lifetime `'a` ties the BasicGenerator to the generator that created it.
+/// The lifetime `'a` ties the BasicGenerate to the generator that created it.
 /// `T: 'a` is required because the parse closure returns `T`.
```
Liam-DeVoe, line 9: this diff is bad, right, this should be kept as BasicGenerator? not sure how this happened

DRMacIver: Yup, that's bad. Will fix.

Liam-DeVoe:
yup not a criticism, I trust you and your changes, just stating a fact about my level of review of this PR!

PR #55, opened 2026-03-12
Support outputting Hegel test results to Antithesis

echoumcp1:


src/antithesis.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub class: String,
+    pub begin_line: u32,
+}
+
+pub(crate) fn emit_assertion(location: &TestLocation, passed: bool) {
+    let output_dir = match std::env::var("ANTITHESIS_OUTPUT_DIR") {
+        Ok(dir) if !dir.is_empty() => dir,
+        _ => return,
+    };
+
```
Liam-DeVoe, line 27: `antithesis.rs` should provide a `is_running_in_antithesis` function, and we use that to guard `emit_assertion` calls. 

`emit_assertion` should also have `if not is_running_in_antithesis: panic!("must be called inside antithesis")`. The reason to have the explicit guard even if we assert is we want to communicate to code readers that they can ignore this function if they are not touching antithesis code.

src/generators/mod.rs
```diff
@@ -14,25 +14,25 @@ pub(crate) mod value;
...
     binary, dates, datetimes, domains, emails, from_regex, ip_addresses, text, times, urls,
 };
 pub use tuples::{
-    tuples10, tuples11, tuples12, tuples2, tuples3, tuples4, tuples5, tuples6, tuples7, tuples8,
-    tuples9,
+    tuples2, tuples3, tuples4, tuples5, tuples6, tuples7, tuples8, tuples9, tuples10, tuples11,
+    tuples12,
 };
 
 pub(crate) use collections::VecGenerator;
```
Liam-DeVoe, line 38: is this reordering manual, or from a linter? If it's from a linter we should figure out why CI hasn't been failing on this before, and either add it as a rule to our CI or decide it's not worth the hassle

tests/test_antithesis.rs
```diff
@@ -0,0 +1,116 @@
...
+    .test_location(hegel::TestLocation {
+        function: "my_test".to_string(),
+        file: "test.rs".to_string(),
+        class: "my_module".to_string(),
+        begin_line: 10,
+    })
+    .run();
+}
+"#;
+
```
Liam-DeVoe, line 19: let's change these tests to use `#[hegel::test]` instead of `hegel::Hegel`. This way we don't need to set `.test_location` manually in the tests.

src/antithesis.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub file: String,
+    pub class: String,
+    pub begin_line: u32,
+}
+
+pub(crate) fn is_running_in_antithesis() -> bool {
+    let output_dir = std::env::var("ANTITHESIS_OUTPUT_DIR");
+    output_dir.is_ok()
+}
+
```
Liam-DeVoe, line 22: Does Antithesis guarantee the dir exists? Should we remove the `output_dir.is_ok()` requirement here and ensure the dir exists before we write to it?

src/generators/mod.rs
echoumcp1: Yes it's from the linter, but it's because I bumped the edition to 2024.

src/antithesis.rs
echoumcp1: It does, but it can't hurt to be extra safe.

Liam-DeVoe: If `ANTITHESIS_OUTPUT_DIR` is set, then we should either (1) error if the directory doesn't exist, or (2) create it if it doesn't. I think it is a mistake to silently treat `ANTITHESIS_OUTPUT_DIR` set to a non-existent directory as not running inside of antithesis, because someone is going to set this var to a non-existent dir and get very confused.

tests/common/project.rs
```diff
@@ -19,7 +25,17 @@ pub struct RunOutput {
...
+        Self::create(main_rs, ProjectMode::Binary)
+    }
+
+    /// Create a temp project with an integration test file (run via `cargo test`).
+    pub fn new_test(test_rs: &str) -> Self {
+        Self::create(test_rs, ProjectMode::Test)
+    }
+
+    fn create(code: &str, mode: ProjectMode) -> Self {
         let temp_dir = TempDir::new().expect("Failed to create temp directory");
```
Liam-DeVoe: I don't think we should make this `mode` distinction. `TempRustProject` should be able to build all reasonable rust projects. `mode` doesn't let you build a project with both a main and a test file, for example.

I would try removing `main_rs` from `TempRustProject.new` and adding two new builder methods, `.main_file` and `.test_file`, where `.main_file` can only be called once.

src/antithesis.rs
echoumcp1: I think creating the directory is reasonable.

Issue #52, opened 2026-03-12
Re-visit `from_type` generator ergonomics and naming


Liam-DeVoe:
See https://github.com/hegeldev/hegel-rust/pull/50#issuecomment-4049029760

PR #57, opened 2026-03-13
Improve default generator ergonomics

DRMacIver:
This makes two big changes:

* Rename `from_type` to `default`, which I think makes a lot more sense in rust. (from_type is a Python name for when you pass an explicit type object. Not a thing in rust)
* Make `default` return a `BoxedGenerator<T>`

The latter is a big ergonomic improvement, because previously the compiler would ~never be able to infer the type parameter to from_type, because it has to figure out not just what T is but what T::Generator is, which requires it to do unbounded search which the rust compiler quite rightly refuses to do.

This does come with one downside, which is that things like `default::<i32>().min_value(0)` don't work, but I think this is fine. This could only have worked in cases where you were passing the type parameter explicitly, and in those cases I think there is very little reason not to just use the generators directly.

Fixes https://github.com/hegeldev/hegel-rust/issues/52

### DIFF
--- a/tests/test_default.rs
+++ b/tests/test_default.rs
@@ -2,6 +2,7 @@ mod common;
 
 use common::utils::check_can_generate_examples;
 use hegel::generators::default;
+use hegel::TestCase;
 use std::collections::HashMap;
 
 #[test]
@@ -88,5 +89,5 @@ fn test_default_can_infer_through_draw(tc: TestCase) {
     // This doesn't test anything much at runtime. We are checking
     // that the type checker can infer the type parameter to default
     // rather than forcing us to write this as default::<i32>
-    let _: i32 = tc.draw(generators::default());
+    let _: i32 = tc.draw(default());
 }


tests/test_compilation_edge_cases.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    let t = tc.draw(refs);
+
+    assert!(t.len() == 3);
+}
+
+#[hegel::test]
+fn default_can_infer_through_draw(tc: TestCase) {
+    let _: i32 = tc.draw(generators::default());
+}
```
Liam-DeVoe: I'd just put this in our `test_from_type.rs` tests (rename to `test_default`?), this isn't really an edge case

tests/test_compilation_edge_cases.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    // Generator for a reference into x, which necessarily
+    // means that `refs` may not outlive `x`.
+    let refs = ix.map(|i| &x[i]).boxed();
+
+    let t = tc.draw(refs);
+
+    assert!(t.len() == 3);
+}
+
```
Liam-DeVoe: I'd also move this to live with our `boxed` tests, I don't really see these as edge cases, they're fundamental choices we made intentionally

tests/test_compilation_edge_cases.rs
DRMacIver: This is the concrete example that doesn't work with the old version of this code FWIW. The thing we're testing is that the compiler can infer that this is a `default::<i32>`.

Happy to put it in from_type though.

Liam-DeVoe: Ah nice. This is clearly an ergonomics win.

PR #58, opened 2026-03-13
Improve getting started guide

rdck:


docs/getting-started.md
```diff
@@ -14,70 +14,39 @@ The SDK requires [`uv`](https://github.com/astral-sh/uv) installed and on your P
...
-    let s = tc.draw(generators::text());
-    assert_eq!(n, n);
-    assert!(s.len() >= 0);
+fn test_bounded_integers_always_below_50(tc: TestCase) {
+    let n = tc.draw(integers::<i32>()
+        .min_value(0)
+        .max_value(49));
+    assert!(n < 50); // this will fail!
 }
 ```
```
Liam-DeVoe, line 45: wait, really? surely integers in [0, 49] are indeed less than 50?

docs/getting-started.md
```diff
@@ -312,6 +281,20 @@ fn test_with_notes(tc: hegel::TestCase) {
...
 
+## Controlling the number of test cases
+By default Hegel runs 100 test cases. Use the builder API to override this:
+
+```rust
```
Liam-DeVoe: this isn't a builder API anymore now that it's on the `#[hegel::test]` macro, should call it "settings" or something

docs/getting-started.md
```diff
@@ -1,82 +1,114 @@
-# Getting Started with Hegel for Rust
+# Getting started with Hegel for Rust
+
+This guide walks you through the basics of installing Hegel and writing your first tests.
+
+## Prerequisites
+
+You will need [`uv`](https://github.com/astral-sh/uv) installed and on your PATH.
```
DRMacIver, line 7: Note to self mostly: We should add some more text here explaining why this is the case.

docs/getting-started.md
```diff
@@ -1,82 +1,114 @@
...
 
 ## Install Hegel
 
-Add Hegel to your `Cargo.toml` as a dev dependency:
+Add `hegel-rust` to your `Cargo.toml` as a dev dependency:
 
 ```toml
 [dev-dependencies]
 hegel = { git = "https://github.com/hegeldev/hegel-rust" }
```
DRMacIver: I think we can now add this as a normal crate dependency, so I guess `hegel = "0.1.0"`

docs/getting-started.md
```diff
@@ -1,82 +1,114 @@
...
-use hegel::generators::{self, Generator};
+use hegel::TestCase;
+use hegel::generators::integers;
 
 #[hegel::test]
-fn test_integers(tc: hegel::TestCase) {
-    let n = tc.draw(generators::integers::<i64>());
-    println!("called with {n}");
-    assert_eq!(n, n); // integers are always equal to themselves
+fn test_integer_self_equality(tc: TestCase) {
```
DRMacIver, line 27: Again mostly a note to self, but really we should have a better example here.

docs/getting-started.md
```diff
@@ -1,82 +1,114 @@
...
-When the test fails, Hegel finds the smallest counterexample — in this case,
-`n = 50`.
+Run the test again. It should now pass.
+```rust
 
-## Generating multiple values
+## Define your own generators
 
-Call `tc.draw()` multiple times to produce multiple values in a single test:
+Hegel provides some generators that you can use out of the box.
```
DRMacIver: I think this is missing a follow on sentence or clause. Something like "You can compose these to generate whatever you want."

docs/getting-started.md
Liam-DeVoe: (publishing to crates.io very very soon! will need `hegeltest = "0.1.0"` though)

docs/getting-started.md
keerlu: Thanks, fixed

docs/getting-started.md
keerlu: Updated

docs/getting-started.md
keerlu: Updated

CONTRIBUTING.md
```diff
@@ -0,0 +1,6 @@
+## Development
```
keerlu, line 1: I moved this here from the README, but if we're going to keep it it should probably have more words

Liam-DeVoe: we can probably just kill this file and delete this content until we have anything more substantial to write here

PR #61, opened 2026-03-13
Add a top level composite generator macro

rdck:
Not quite ready to be merged yet.

TODO:
+ Passthrough parameters
+ Better error messages when you hold it wrong

### DIFF
--- a/tests/test_composite.rs
+++ b/tests/test_composite.rs
@@ -83,10 +83,8 @@ fn main() {}
     let output = TempRustProject::new(code).run();
     assert!(!output.status.success());
     assert!(
-        output
-            .stderr
-            .contains("must define a first parameter of type TestCase"),
-        "Expected missing return type error, got: {}",
+        output.stderr.contains("must have type TestCase"),
+        "Expected TestCase parameter error, got: {}",
         output.stderr
     );
 }


hegel-derive/src/composite.rs
```diff
@@ -0,0 +1,71 @@
...
+    };
+    let tc_pattern = &tc_arg.pat;
+    let tc_type = &tc_arg.ty;
+
+    // Check if the type path for the first parameter ends with a "TestCase" identifier. As far as
+    // I can tell, there's no good way of checking at macro expansion time whether this actually
+    // resolves to the right type.
+    let Type::Path(path) = tc_type.as_ref() else {
+        panic!("{}", MISSING_TEST_CASE_PARAMETER)
+    };
```
Liam-DeVoe: Hmm. AIUI this check can have false alarms if a user renames `TestCase` during import:

`use hegel::TestCase as Blarg`

Given this, and that if someone leaves the `TestCase` parameter out they're still going to get a compile-time error from rust expanding the macro, I'm wondering if it's worth having these proactive errors. Ideally rust would have some way to say "if the macro expansion fails, run this code that has some heuristics to provide better error messages", which would be the right place for this to run.

rdck: Good point.

If we remove the check, the error you get isn't terrible:
```
error[E0425]: cannot find value `tc` in this scope
```

> Ideally rust would have some way to say "if the macro expansion fails, run this code that has some heuristics to provide better error messages", which would be the right place for this to run.

Yeah. AFAIU, because macro expansion happens before compilation, the macro can't inspect the result of compilation.

rdck: I say let's just remove the check.

Liam-DeVoe: I'm also in favor of removing the check 👍

rdck: Actually, I'm realizing the error message can be a lot more confusing, if the generator doesn't actually ever use the test case object. I expect that would be relatively rare, but still less than ideal.

Liam-DeVoe: Let's:

(1) Add a macro-time error for the case when the function has zero arguments, which we can detect precisely with no false alarms
(2) Do the following, suggested by Claude after a conversation and sounds perfect to me:

```
  Case 2 (wrong first param type): Don't try to detect it in the macro. Instead, define a marker trait with #[diagnostic::on_unimplemented]:

  #[diagnostic::on_unimplemented(
      message = "`#[composite]` requires `&mut TestCaseData` as the first parameter",
      label = "this type does not match `&mut TestCaseData`",
      note = "example: `#[composite] fn my_gen(data: &mut TestCaseData, ...) -> T`"
  )]
  pub trait IsTestCaseData {}

  impl IsTestCaseData for TestCaseData {}

  Then in the macro's expansion, emit a bound like:

  fn _assert_data<T: IsTestCaseData>() {}
  _assert_data::<FirstParamType>();

  where FirstParamType is extracted from the first parameter's type annotation (strip the &mut).
```

src/test_case.rs
```diff
@@ -9,6 +9,14 @@ use std::sync::{Arc, LazyLock};
 
 use crate::generators::value;
 
+// This is used only as a mechanism for printing a nice error message when the user forgets the
+// test case parameter.
+#[diagnostic::on_unimplemented(
+    message = "Functions marked with #[composite] must have a first parameter of type TestCase.",
+    label = "This type does not match `TestCase`."
```
Liam-DeVoe: We should standardize on whether we use `#[hegel::composite]` or `#[composite]` in our docs, our tests, and our user-facing examples. This will become the de-facto recommended style to import hegel, so it's important it's a style we're happy with seeing users use. 

Similar to how we consistently use `st.integers()` in hypothesis.

@DRMacIver opinions?

Liam-DeVoe: See also: https://github.com/hegeldev/hegel-rust/issues/75

PR #65, opened 2026-03-15
Stateful testing

rdck:
Here's a proposal for the stateful testing interface.

### DIFF
--- a/examples/ledger.rs
+++ b/examples/ledger.rs
@@ -13,6 +13,12 @@ struct Ledger {
 }
 
 impl Ledger {
+    fn new() -> Self {
+        Ledger {
+            balances: HashMap::new(),
+        }
+    }
+
     fn credit(&mut self, account: String, amount: i64) {
         let balance = self.balances.entry(account).or_insert(0);
         *balance += amount;
@@ -29,12 +35,6 @@ impl Ledger {
         if from != to && amount - from_balance <= 1 {
             self.debit(from, amount);
             self.credit(to, amount);
-        }
-    }
-
-    fn new() -> Ledger {
-        Ledger {
-            balances: HashMap::new(),
         }
     }
 }
--- a/examples/min_stack.rs
+++ b/examples/min_stack.rs
@@ -18,7 +18,7 @@ struct MinStack<T> {
 }
 
 impl<T: Copy + Ord> MinStack<T> {
-    fn new() -> MinStack<T> {
+    fn new() -> Self {
         MinStack {
             stack: Vec::new(),
             minimums: Vec::new(),


examples/die_hard.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+#[hegel::test(test_cases = 2000)]
+fn test_die_hard(tc: TestCase) {
+    let m = DieHard { small: 0, big: 0 };
+    hegel::stateful::run(m, tc);
+}
+
+fn main() {}
```
Liam-DeVoe, line 66: @DRMacIver one thing I think is worse in this design from Hypothesis is a stateful test no longer has a 1:1 correspondence with an `@given` test. In this PR, we define a normal `#[hegel::test]`, and just happen to call `hegel::stateful::run(m, tc)` inside of it. There's nothing stopping me from drawing some unrelated value and making an unrelated assertion. Heck I could run `hegel::stateful::run(m, tc)` twice if I wanted!

Nothing *breaks*, but this does muddy the conceptual model around eg stateful-specific settings like `max_stateful_steps`.

It's also not super obvious `hegel::stateful::run(m, tc)` runs a single test case and not the entire 100 test cases of the state machine, though I think people coming from not-hypothesis won't feel this as strongly.

src/stateful.rs
```diff
@@ -0,0 +1,155 @@
...
+    let step_cap = min(unbounded_step_cap, max_steps);
+
+    let mut steps_run_successfully = 0;
+    let mut steps_attempted = 0;
+
+    while steps_run_successfully < step_cap
+        && (steps_attempted < 10 * step_cap
+            || (steps_run_successfully == 0 && steps_attempted < 1000))
+    {
+        let rule = tc.draw(rules);
```
Liam-DeVoe: I was going to leave a comment of "I am very skeptical of this `< 1000` condition" but apparently this is what Hypothesis does!

examples/die_hard.rs
DRMacIver: I'm actually sortof in favour of being able to run unrelated assertions and even multiple stateful tests in a single hegel test. I don't have a really good concrete use case for this, but I have an intuition that it will be powerful.

src/stateful.rs
DRMacIver: Yeah it's not a perfect set of heuristics, but it's mostly good enough. It's basically just an infinite loop catcher.

src/stateful.rs
```diff
@@ -0,0 +1,155 @@
...
+pub type Rule = Box<dyn FnMut()>;
+
+pub struct Variables<T> {
+    pool_id: i128,
+    tc: TestCase,
+    values: HashMap<i128, T>,
+}
+
+impl<T> Variables<T> {
+    fn pool_generate(&self, consume: bool) -> i128 {
```
DRMacIver, line 45: I think this is fine as is for now, but I'd like to register a desire to improve this interface later.

The main thing is that I think that this isn't actually a boolean flag, because the big difference between consuming and non-consuming draws in rust is that the latter can be references while the former have to be values.

Also something I'd like and that we don't currently have here is that there should be generators corresponding to a variables instance. Something like `references_from` and `values_from` maybe.

But none of this should block merging this, which I think is good enough for a first release.

examples/die_hard.rs
Liam-DeVoe: yeah I can see it. I do think the confusion with test settings is real though, although I don't have quite the right words to express why

src/test_case.rs
```diff
@@ -168,41 +190,59 @@ impl TestCase {
...
+        TestCase {
+            global: self.global.clone(),
+            local: RefCell::new(TestCaseLocalData {
+                span_depth: 0,
+                draw_count: 0,
+                indent: local.indent + extra_indent,
+                on_draw: local.on_draw.clone(),
+            }),
         }
     }
```
Liam-DeVoe, line 218: not a huge fan of this, feels like we're going to use this capability exactly once ever (for stateful testing and nothing else) and requiring us to reason about arbitrary nestings of global and local test cases datas is going to be painful for reasoning about code correctness. I'd prefer a `stateful_depth_start` or similar on `TestCaseData`, accepting that we'll have some hardcoded things for stateful testing.

 Not blocking for release but I want to think about a different design here

examples/die_hard.rs
DRMacIver: I think having `max_stateful_steps` in settings is a bad idea that I've wanted to remove for a while in Hypothesis tbh. It's currently hard-coded to 50 here, which seems reasonable to me, though it would be nice to make it more configurable.

src/test_case.rs
DRMacIver: I think we want it in other places where we capture a TestCase object actually. For example, I was thinking that we wanted something like this for having reporting in our fake randoms objects.

DRMacIver: We also need to be able to do something like this once we start e.g. wanting to be able to fork test case objects to run in multiple threads. We're a long way from that working, just highlighting that it will be needed eventually.

Liam-DeVoe: hmm ok. This is going to have rather large implications for our invariants around spans and printing and I want to think hard about this before we say we're both happy with it. Not discarding it out of hand but this is actually a rather big internal API decision

DRMacIver: Happy to talk about it more, but I propose that getting this right not be release blocking as it's internal API only.

Liam-DeVoe: yes definitely this is for later

PR #82, opened 2026-03-19
Handle server crashes gracefully instead of hanging

DRMacIver:
When the server crashes we would previously end up hanging. This sucks. This PR makes it so that we instead raise an error when that happens.


Liam-DeVoe:
nice 👍

PR #88, opened 2026-03-19
Emit drawn values and seed in Antithesis

echoumcp1:


src/runner.rs
```diff
@@ -721,22 +721,25 @@ where
...
     test_channel: Channel,
     test_fn: &mut F,
     is_final: bool,
     verbosity: Verbosity,
     got_interesting: &Arc<AtomicBool>,
-) {
+) -> Vec<String> {
     // Create TestCase. The test function gets a clone (cheap Rc bump),
     // so we retain access to the same underlying TestCaseData after the test runs.
     let tc = TestCase::new(Arc::clone(connection), test_channel, verbosity, is_final);
```
Liam-DeVoe, line 736: this is a brittle return value. We want to return something like Hypothesis's `ConjectureResult` instead. (I'm also not 100% sure `run_test_case` is the right place to be returning anything from and would need to poke through the code)

tests/test_antithesis.rs
```diff
@@ -76,11 +77,69 @@ fn my_test(tc: hegel::TestCase) {
...
+    let code = r#"
+use hegel::generators;
+
+#[hegel::test]
+fn my_test(tc: hegel::TestCase) {
+    let _x: bool = tc.draw(generators::booleans());
+    let _y: bool = tc.draw(generators::booleans());
+    panic!("always fails");
+}
+"#;
```
Liam-DeVoe, line 100: no need to assign these to a var at all if they're unused, it reads more cleanly that way

echoumcp1: claudism 😅

src/runner.rs
echoumcp1: makes sense this was mostly just to see if it would display in Antithesis reasonably

tests/test_antithesis.rs
DRMacIver: Claude fuckin' loves underscores.

PR #89, opened 2026-03-20
Drop `num` dependency, add `test-minimal` CI job

Liam-DeVoe:
~~`#![cfg_attr(docsrs, feature(doc_cfg))]` is the thing that actually fixes the docs build errors we've been getting: https://docs.rs/crate/hegeltest/0.1.9/builds/3037114.~~

~~I changed our docs ci to use `cargo-docs-rs`, which would have caught this.~~

The new `test-minimal` ci job is unrelated but is a good thing to have. This tests against the minimal declared versions of all our dependencies. We should add this in Hypothesis, tbh. `uv` and probably others supports it: https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -883,9 +883,7 @@ fn run_test_case<F: FnMut(TestCase)>(
     // so we retain access to the same underlying TestCaseData after the test runs.
     let tc = TestCase::new(Arc::clone(connection), test_channel, verbosity, is_final);
 
-    let result = with_test_context(|| {
-        catch_unwind(AssertUnwindSafe(|| test_fn(tc.clone())))
-    });
+    let result = with_test_context(|| catch_unwind(AssertUnwindSafe(|| test_fn(tc.clone()))));
 
     let (status, origin) = match &result {
         Ok(()) => ("VALID".to_string(), None),


tests/common/project.rs
```diff
@@ -3,8 +3,13 @@
 
 use std::path::PathBuf;
 use std::process::{Command, ExitStatus};
+use std::sync::atomic::{AtomicU64, Ordering};
 use tempfile::TempDir;
 
+// use a unique package name in our Cargo.toml to avoid cargo using the cached build from
+// a different test
+static PACKAGE_NAME_ID: AtomicU64 = AtomicU64::new(0);
```
DRMacIver, line 24: I'm not sure I understand what's going on here, but is it enough to have this be unique within a given cargo  test run? What if it was used in previous test runs?

Liam-DeVoe: good point. I think my comment here was just wrong. Parallel cargo builds hold a file lock and so I think the real reason to do this is avoiding lock contention. 

The reuse across previous test runs makes me quite nervous so we now wipe the cache dir at the start of each test suite.

Liam-DeVoe:
I've also gotten rid of our dependency on `num` in this PR since it's ~trivial to roll our own generic `Integer` trait.

src/generators/numeric.rs
```diff
@@ -1,9 +1,41 @@
 use super::{BasicGenerator, Generator, TestCase};
 use crate::cbor_utils::{cbor_map, cbor_serialize, map_insert};
 use ciborium::Value;
-use num::{Bounded, Float as NumFloat, Integer as NumInteger};
 use std::marker::PhantomData;
 
+pub trait Integer: Copy + Ord {
```
DRMacIver, line 6: I'm happy to go with your judgment on this but registering that I don't love reimplementing integer traits ourselves and think a `num` dependency is probably fine.

src/cbor_utils.rs
```diff
@@ -41,10 +39,11 @@ pub fn map_get<'a>(value: &'a Value, key: &str) -> Option<&'a Value> {
         panic!("expected Value::Map, got {value:?}");
     };
     for (k, v) in entries {
-        if let Value::Text(s) = k {
-            if s == key {
-                return Some(v);
-            }
+        let Value::Text(s) = k else {
```
DRMacIver, line 42: Huh I'd not seen `let ... else ...` syntax before. Neat.

src/control.rs
```diff
@@ -1,12 +1,16 @@
-use std::cell::Cell;
+use std::cell::RefCell;
+
+use crate::test_case::TestCase;
 
 thread_local! {
-    static IN_TEST_CONTEXT: Cell<bool> = const { Cell::new(false) };
+    static CURRENT_TEST_CONTEXT: RefCell<Option<TestCase>> = const { RefCell::new(None) };
```
DRMacIver: I'm sortof keen to not have us depend on a global testcase object. It tends to get pretty messy over time as more things depend on it rather than the explicit test case object they've got access to. Can you say more about why this was needed?

tests/common/utils.rs
```diff
@@ -9,6 +9,15 @@ use hegel::{Hegel, Settings};
 use regex::Regex;
 use std::fmt::Debug;
 
+// some of our tests differ in behavior in our nightly rust job.
+pub fn is_nightly() -> bool {
+    let output = std::process::Command::new("rustc")
```
DRMacIver: Is this really the best way to check if we're on nightly?? Surely not. We want to know if we were compiled with nightly, while this checks if the current environment is configured to use nightly.

src/generators/numeric.rs
Liam-DeVoe: I think this is the right general (internal) API design for Hegel. We're going to have a bunch of strategies which want to accept various arguments and are going to want semi fine grained control over this. For example we're eventually going to want to support `integers` generating num's `BigInt`. If we used `num` as our trait, this would be totally fine. But tomorrow some competitor `awesome-num` is going to come along with their own `BigInt` that we also want to support generating! Now we're stuck and tied down to `num`. I think having our own traits avoids a bunch of headaches.

Similar opinions for having internal traits for eg `from_regex`, see https://github.com/hegeldev/hegel-rust/issues/92

src/control.rs
Liam-DeVoe: don't think this is used by anything right now, this was future-proofing. 

I basically view this as dependency injection. One could thread through the explicit parameters everywhere, but the hypothesis engine has like 5 layers of control flow and it just gets too much. We're very aggressive about limiting the fields that get to go on the build context and who accesses them and always prefer passing through explicit params, only using the test case context when it would be overly onerous to do so.

Hegel libraries are a lot simpler than the hypothesis engine so this may not be required but I kind of suspect it will be.

tests/common/utils.rs
Liam-DeVoe: lol yeah this is terrible. I'll look around

src/generators/numeric.rs
Liam-DeVoe: also this lets us drop an actually relatively heavy dep and you know how much I love doing that :)

(we could have improved this substantially by only pulling in `num-traits` instead of the umbrella `num`, but)

tests/common/utils.rs
Liam-DeVoe: I think there actually might not be a good way to do this. I'm sidestepping the issue by setting an envvar in CI

src/control.rs
Liam-DeVoe: have reverted to a bool for now, can revisit later

PR #95, opened 2026-03-23
.github/scripts/release.py: also update conformance Cargo.lock

winterqt:
Avoids a bunch of "Updated hegel..." on conformance test run because we bumped the version pulled in via path dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,6 @@ exclude = [
   "/.claude",
   "/.github",
   "/nix",
-  "/tests/conformance",
-  "/justfile",
   "/RELEASE-sample.md",
 ]
 license = "MIT"
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -139,9 +139,13 @@ def release() -> None:
     set_version(ROOT / "hegel-macros" / "Cargo.toml", new_version)
     set_macros_dep_version(ROOT / "Cargo.toml", new_version)
 
-    # regenerate lockfile after version bump
+    # regenerate lockfiles after version bump
     subprocess.run(["cargo", "generate-lockfile"], check=True, cwd=ROOT)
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -7,6 +7,14 @@ version = "0.1.14"
...
 rust-version = "1.86"
+exclude = [
+  "/.claude",
+  "/.github",
+  "/nix",
+  "/tests/conformance",
+  "/justfile",
+  "/RELEASE-sample.md",
+]
 license = "MIT"
```
Liam-DeVoe, line 16: hmm, should we use `include` instead?

winterqt: I'm not *opposed* but it also might be a bit harder to keep up with? I don't expect this list to change drastically over time tbh.

winterqt: I guess for includes we could just do `/docs`, `/src`, `/tests`, `/LICENSE`, and `/README.md` or whatever else obvious that I'm missing. I personally like the exclude spelling better myself, but shrug.

Liam-DeVoe: `exclude` is harder to keep up with and more brittle than `include`!

Liam-DeVoe: I thought about it more and a `include` makes me more nervous about accidentally forgetting to include a file and publishing a broken version than `exclude` does. I'm going with a trimmed-down `exclude` here

winterqt: Why include `/tests/conformance` and `/justfile`? The former is definitely not necessary and will always arbitrarily change on every release because of the lockfile. :(

(I care less about `justfile` being there.)

Liam-DeVoe: we're already including our tests, and the conformance tests are part of our tests! I don't see why we would exclude the conformance tests but not the rest of the tests. 

Is there a reason to be concerned about including files which change on every release?

PR #97, opened 2026-03-23
Nix fixes

winterqt:
Depends on https://github.com/hegeldev/hegel-core/pull/62.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 **/*.rs.bk
 .idea/
 .vscode/
+.direnv/
+.envrc
 *.swp
 *.swo
 .hypothesis/


nix/flake.nix
```diff
@@ -33,16 +26,14 @@
...
               pkgs.cargo
               pkgs.rustc
               pkgs.rustfmt
               pkgs.clippy
               pkgs.just
-              hegel.packages.${system}.default
             ];
-            HEGEL_SERVER_COMMAND = "${hegel.packages.${system}.default}/bin/hegel";
+            HEGEL_SERVER_COMMAND = pkgs.lib.getExe hegel.packages.${system}.default;
           };
```
Liam-DeVoe, line 38: I assume you tested this? I thought we discussed this and decided it wouldn't work because `/bin/hegel` is a weird nonstandard location

winterqt: Yes, it works just fine -- `/bin/{name}` is the standard location for executables; I'm afraid I don't remember that conversation.

PR #106, opened 2026-03-24
Update docs with better hegel install instructions

DRMacIver:


src/lib.rs
```diff
@@ -10,11 +10,10 @@
...
 //! ## Install Hegel
 //!
-//! Add `hegel-rust` to your `Cargo.toml` as a dev dependency:
+//! Add `hegel-rust` to your `Cargo.toml` as a dev dependency using cargo:
 //!
-//! ```toml
-//! [dev-dependencies]
-//! hegeltest = "0.1.0"
+//! ```bash
+//! cargo add --dev hegeltest
```
Liam-DeVoe, line 16: non-blocking but I'd drop `using cargo` - what else would they be using!

DRMacIver: I was imagining contrasting it with a text editor.

PR #109, opened 2026-03-24
Some documentation fixes

winterqt:


Cargo.toml
```diff
@@ -26,6 +26,11 @@ categories = ["development-tools::testing"]
 [lib]
 name = "hegel"
 
+# Required for rustdoc to know that we're okay with building examples.
+[[example]]
+name = "stack"
+doc-scrape-examples = true
+
 [dependencies]
```
Liam-DeVoe, line 37: this is listing just one of the examples, `stack.rs`. Is it still going to scrape all of the other example files?

Liam-DeVoe: I don't see the requirement that we list `[[examples]]` in cargo.toml mentioned in the docs: https://doc.rust-lang.org/rustdoc/scraped-examples.html

Cargo seems to autodiscover the examples fine:

```
 ~/Desktop/Liam/coding/hegel/rust λ cargo build --example
error: "--example" takes one argument.
Available examples:
    die_hard
    ledger
    min_stack
    stack

 ~/Desktop/Liam/coding/hegel/rust λ 
```

Liam-DeVoe: I went down this rabbit hole, and it is required, but for a more specific reason than the comment implied. Here's the docs page that convinced me: https://doc.rust-lang.org/cargo/reference/unstable.html#scrape-examples

src/generators/mod.rs
```diff
@@ -59,5 +59,4 @@ pub use tuples::{
 };
 
 #[cfg(feature = "rand")]
-#[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
 pub use random::{HegelRandom, RandomsGenerator, randoms};
```
Liam-DeVoe, line 62: I confirmed this removal is correct, but only on nightly: https://github.com/rust-lang/rust/issues/43781. docs-rs builds on nightly so this is fine

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch improves documentation and adds citations to examples.
```
```suggestion
RELEASE_TYPE: patch

This patch improves documentation and adds scraped examples to the docs.
```

PR #112, opened 2026-03-24
.github/workflows/ci: remove no longer needed deploy key usage

winterqt:


winterqt:
We probably actually want to just move the `bump-hegel-core` and `release` jobs to a separate environment like Zizmor recommends... no idea how easy that is to do from the GitHub side, though.

PR #113, opened 2026-03-25
Use stdio instead of Unix socket

DRMacIver:
Switches server communication from Unix sockets to stdin/stdout pipes, using hegel-core's new `--stdio` flag. Removes socket path, temp dir, and connect-retry logic.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,4 @@
 RELEASE_TYPE: patch
 
-This release moves the rust library over to using the new stdio version of hegel-core.
-
+This release moves over to using the new stdio version of hegel-core.
 This should not be a user visible change.


.gitignore
```diff
@@ -21,3 +21,4 @@ result
 .claude/*.local.json
 .claude/*.local
 .claude-reliability/
+hegel-core
```
Liam-DeVoe: ? should delete this

DRMacIver: Yup. Relic of some experiments, sorry..

RELEASE.md
```diff
@@ -0,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This release moves the rust library over to using the new stdio version of hegel-core.
+
+This should not be a user visible change.
```
Liam-DeVoe, line 4: should drop any explicit references to the "rust library" - they're already reading the changelog for the rust library!

More generally, changelogs written with phrases like "the rust library" make people reading them feel like we aren't treating them as first class citizens

DRMacIver: "written like this"?

Liam-DeVoe: yeah sorry that was ambiguous, edited!

Liam-DeVoe: in particular I didn't mean to imply that I had a problem with any of the changelog except for the "rust library" phrase

PR #114, opened 2026-03-25
Publish to crates.io before pushing to main, fall back to PR on failure

DRMacIver:
## Summary

- The release script previously pushed the tag and main branch in a single `git push origin main --tags`. If main had diverged (e.g. a PR merged during CI), both the push and the subsequent crates.io publish would fail — which is what happened in [this run](https://github.com/hegeldev/hegel-rust/actions/runs/23505886841/job/68413938573).
- Now the tag is pushed first (separately), then crates.io publish happens, and only then does it try pushing to main. If the main push fails, it opens a PR with a `skip release` label to merge the release commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Check `cargo package`
         # creates a .crate for distribution and verifies it can build in isolation
         # see https://doc.rust-lang.org/cargo/commands/cargo-package.html
-        run: cargo package
+        run: cargo package --workspace
 
   test:
     name: test (${{ matrix.rust }})


.github/scripts/release.py
```diff
@@ -195,6 +195,52 @@ def release() -> None:
...
+
+    # Ensure the "skip release" label exists so check-release doesn't run on this PR
+    subprocess.run(
+        [
+            "gh", "label", "create", "skip release",
+            "--force",
+            "--description", "Skip the release check on this PR",
+        ],
+        cwd=ROOT,
+    )
```
Liam-DeVoe, line 224: we should *not* be force creating labels. This workflow needs to fail if the label doesn't exist

DRMacIver: True, sorry. Missed that on my review. Also sorry to merge this without your review, but build debugging nonsense.

Liam-DeVoe: yeah understood 👍

PR #115, opened 2026-03-25
Use a persistent hegel server session across all tests

DRMacIver:
Previously, every Hegel::run() call spawned a new Python subprocess, created a Unix socket, performed a handshake, ran one test, tore down the connection, and waited for the child to exit. This was never the intended model, but somehow snuck in (I presume Claude making a "pragmatic decision" and me failing to notice). This moves us over to the actually intended model of having a single persistent 

Now a single HegelSession is created on first use and reused for all `Hegel::run()` calls within the process. 

This also moves the Connection over to use a background reader thread that dispatches incoming packets to per-channe queues, matching the Python server's architecture. This allows multiple tests to run concurrently over the same connection without stream lock contention. Only the control channel (run_test send/receive) is serialized via a Mutex.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -49,9 +49,10 @@ static PANIC_HOOK_INIT: Once = Once::new();
 struct HegelSession {
     _temp_dir: TempDir,
     connection: Arc<Connection>,
-    /// Serializes all test execution. The protocol's stream-level locking
-    /// doesn't support concurrent readers on different channels, so only
-    /// one test can use the connection at a time.
+    /// The control channel is shared across threads, so it's behind a Mutex
+    /// because Channel is not thread-safe. The lock is only held for the
+    /// brief run_test send/receive; test execution runs concurrently on
+    /// per-test channels.
     control: Mutex<Channel>,
 }
 


src/runner.rs
```diff
@@ -333,16 +468,7 @@ pub enum Verbosity {
...
-            Verbosity::Quiet => "quiet",
-            Verbosity::Normal => "normal",
-            Verbosity::Verbose => "verbose",
-            Verbosity::Debug => "debug",
-        }
-    }
-}
+impl Verbosity {}
 
 // internal use only
```
Liam-DeVoe, line 347: intentional deletion? I liked these debug strings

src/runner.rs
```diff
@@ -689,14 +697,20 @@ where
...
+            let mut control = session.control.lock().unwrap();
+            run_test_id = control
+                .send_request(cbor_encode(&run_test_msg))
+                .expect("Failed to send run_test");
+
+            let run_test_response = control
+                .receive_reply(run_test_id)
+                .expect("Failed to receive run_test response");
+            let _run_test_result: Value = cbor_decode(&run_test_response);
+        }
```
Liam-DeVoe, line 715: I'm confused why this serialization is necessary. Didn't we design the protocol so it's never ambiguous which packet is being responded to?

src/runner.rs
```diff
@@ -834,20 +833,9 @@ where
...
-        let _ = monitor_handle.join();
-
         let test_failed = !passed || got_interesting.load(Ordering::SeqCst);
 
         if is_running_in_antithesis() {
-            // if we're running inside of antithesis, but the user hasn't opted in
-            // to the antithesis feature, loudly inform them.
             #[cfg(not(feature = "antithesis"))]
             panic!(
                 "When Hegel is run inside of Antithesis, it requires the `antithesis` feature. \
```
Liam-DeVoe, line 842: intentional deletion or claude deletion? I guess this comment doesn't add much to the code below it, so I'm not super opposed to removing it

src/runner.rs
DRMacIver: Intentional only in the sense that they weren't used anywhere now. I think we need to think harder about what the verbosity settings actually do at some point, because I think they're not actually useful to pass back to the server.

src/runner.rs
DRMacIver: Yeah, I too am confused. This looks wrong to me.

src/runner.rs
Liam-DeVoe: Yeah happy to have that conversation. I do still think it's useful to pass back, because verbosity=debug enables `HEGEL_PROTOCOL_DEBUG` on the server.

src/runner.rs
DRMacIver: Claude deletion, don't know why, but I'm also not against deleting it.

src/runner.rs
DRMacIver: Actually I think the comment is wrong but the code is correct. The channel data structure isn't thread safe so we need to lock to use the control channel to ask to start running a test. The protocol could handle multiple message sends on it no problem.

This is a fairly fast call and response though, it's not the code for running the test, just to start running a new test, so I think it's OK to lock.

Liam-DeVoe: ok great, so it's a rust-level limitation, makes sense. let's just update the comment then

src/runner.rs
DRMacIver: I thought we decided that it couldn't do that because you needed to set that on a per-connection rather than per-test basis earlier in the process than debug has access to?

src/runner.rs
DRMacIver: Yup, done.

src/runner.rs
Liam-DeVoe: I think that's true today but I'd ideally like tests that set verbosity to also carry that over to the server, which tells the server to log messages related to just that test, as if `HEGEL_PROTOCOL_DEBUG` were set process-wide

DRMacIver: I'm certainly not against supporting that. I think it will be tricky and of somewhat limited utility, but it's certainly not harmful to support.

PR #124, opened 2026-03-26
Fix hang when HEGEL_SERVER_COMMAND points to incompatible binary

DRMacIver:
Tests would hang if you were using an old version of hegel-core that didn't support the `--stdio` flag. This fixes that and adds some comprehensive debugging messages when the server start doesn't work.


src/runner.rs
```diff
@@ -954,3 +1095,128 @@ fn cbor_encode(value: &Value) -> Vec<u8> {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_wait_for_exit_child_exits() {
+        let mut child = Command::new("true").spawn().unwrap();
+        let result = wait_for_exit(&mut child, Duration::from_secs(5));
+        assert!(result.is_some());
+    }
+
```
Liam-DeVoe: can these tests be moved to a separate file? I want to keep our source code readable

src/runner.rs
```diff
@@ -344,28 +348,165 @@ fn ensure_hegel_installed() -> Result<String, String> {
...
+    match Command::new(binary_path).arg("--version").output() {
+        Ok(output) if output.status.success() => {
+            let stdout = String::from_utf8_lossy(&output.stdout).trim().to_string();
+            if stdout != expected_version_string {
+                parts.push(format!(
+                    "Version mismatch: expected '{}', got '{}'.",
+                    expected_version_string, stdout
+                ));
+            }
+        }
```
Liam-DeVoe: If we ever change the hegel `--version` format (which we will, because it's not part of the standard, and also https://github.com/hegeldev/hegel-core/issues/67), users are going to get a misleading error message here. 

This logic should be robust to that happening. Either simply include the output of `--version` in the message as "hey it might be helpful for you to know that `hegel --version` returned this", or standardize `--version` as part of the protocol. Either is fine with me.

src/runner.rs
```diff
@@ -344,28 +348,165 @@ fn ensure_hegel_installed() -> Result<String, String> {
...
+        if let Ok(path_var) = std::env::var("PATH") {
+            for dir in std::env::split_paths(&path_var) {
+                let candidate = dir.join(path);
+                if candidate.is_file() {
+                    let resolved = candidate.to_string_lossy().to_string();
+                    validate_executable(&resolved);
+                    return resolved;
+                }
+            }
+        }
```
Liam-DeVoe, line 396: "any sufficiently complicated rust project contains an ad hoc, informally-specified, bug-ridden, slow implementation of half of the python standard library" ...

I miss `shutil.which()` :(

src/runner.rs
```diff
@@ -344,28 +348,165 @@ fn ensure_hegel_installed() -> Result<String, String> {
...
+    // Bare name (no '/') — try PATH lookup
+    if !path.contains('/') {
+        if let Ok(path_var) = std::env::var("PATH") {
+            for dir in std::env::split_paths(&path_var) {
+                let candidate = dir.join(path);
+                if candidate.is_file() {
+                    let resolved = candidate.to_string_lossy().to_string();
+                    validate_executable(&resolved);
+                    return resolved;
+                }
```
Liam-DeVoe: I would extract this to `fn which` in `utils.rs`

src/runner.rs
```diff
@@ -344,28 +348,165 @@ fn ensure_hegel_installed() -> Result<String, String> {
...
+            if metadata.permissions().mode() & 0o111 == 0 {
+                panic!(
+                    "Hegel server binary at '{}' is not executable. \
+                     Check file permissions.",
+                    path
+                );
+            }
+        }
+    }
+}
```
Liam-DeVoe: I would move this to `utils.rs`

src/runner.rs
DRMacIver: I wonder if there's a rust port of the Python standard library.

tests/test_bad_server_command.rs
```diff
@@ -0,0 +1,131 @@
...
+fn test_bare_name_not_on_path_gives_informative_error() {
+    TempRustProject::new()
+        .main_file(HEGEL_CODE)
+        .env("HEGEL_SERVER_COMMAND", "definitely_not_a_real_hegel_binary")
+        .expect_failure("not found on PATH")
+        .cargo_run(&[]);
+}
+
+#[test]
+#[cfg(unix)]
```
DRMacIver, line 69: Claude added these and they're a bit redundant in that it's not like we support anything that isn't unix right now, but they're not wrong so I decided to leave them.

PR #125, opened 2026-03-26
Give invariants permission to mutate machine state

rdck:
As discussed offline.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,15 @@
 RELEASE_TYPE: minor
 
-This revision removes the type-level restriction on invariants mutating machine state.
+This release changes `self` in `#[invariant]` from an immutable reference to a mutable reference:
+
+```rust
+# before
+#[invariant]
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This revision removes the type-level restriction on invariants mutating machine state.
```
````suggestion
RELEASE_TYPE: minor

This release changes `self` in `#[invariant]` from an immutable reference to a mutable reference:

```rust
# before
#[invariant]
fn my_invariant(&self, ...) {} 

# after
#[invariant]
fn my_invariant(&mut self, ...) {}
```

This will require updating your invariant signatures, but should be strictly more expressive.
````

This is a breaking release, so we should put more care into the release notes than normal.

DRMacIver:
> @rdck or @DRMacIver feel free to merge at will

BTW @rdck can no longer merge because we removed him from the team as part of the team-minimalism effort, which I think we should revert.

PR #129, opened 2026-03-26
Move hegel-core install to XDG cache dir

### LINKED ISSUE #108: `.hegel` venv issues with containerized development
Installed and ran hegel in a container, but then trying to run the tests outside the container results in this error (and vice versa):

```
---- vec_structs_roundtrip stdout ----

thread 'vec_structs_roundtrip' (31955226) panicked at .../.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hegeltest-0.2.0/src/runner.rs:557:14:
Failed to spawn hegel at path .hegel/venv/bin/hegel: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

DRMacIver:
Install hegel-core into $XDG_CACHE_HOME/hegel/versions/<version>/venv (or a suitable fallback for XDG_CACHE_HOME if not set) instead of the project-local .hegel/venv 

Fixes #108 (more or less. Technically one can still trigger the bad behaviour if you work really hard, but it requires you to be doing something really screwy)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+Move default hegel install to a versioned directory in $XDG_CACHE_HOME/hegel rather than .hegel/venv.
+
+This fixes problems with workflows that share a project directory with a container and then end up broken in either the host or container because the virtualenv points to a python that doesn't exist.


src/runner.rs
```diff
@@ -931,3 +1022,556 @@ fn cbor_encode(value: &Value) -> Vec<u8> {
...
+    use super::*;
+    use std::sync::Mutex as StdMutex;
+    use tempfile::TempDir;
+
+    // Environment variable tests must run serially since env vars are process-global.
+    // Use into_ok() to recover from poison (the should_panic test poisons the mutex).
+    static ENV_LOCK: StdMutex<()> = StdMutex::new(());
+
+    fn lock_env() -> std::sync::MutexGuard<'static, ()> {
+        ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner())
```
Liam-DeVoe, line 1037: these tests should live in a tests file to keep the actual code file clean

src/runner.rs
```diff
@@ -9,14 +9,72 @@ use std::backtrace::{Backtrace, BacktraceStatus};
...
+const FILE_LOCK_TIMEOUT: Duration = Duration::from_secs(300);
+const FILE_LOCK_POLL_INTERVAL: Duration = Duration::from_millis(500);
+
+/// Returns the cache directory for hegel installations.
+///
+/// Resolution order:
+/// 1. `$XDG_CACHE_HOME/hegel` if `XDG_CACHE_HOME` is set
+/// 2. `~/Library/Caches/hegel` on macOS
+/// 3. `~/.cache/hegel` on other platforms
+fn hegel_cache_dir() -> PathBuf {
```
Liam-DeVoe, line 30: I'd actually prefer `~/.cache/hegel` even on macos. Yes, this is strictly speaking not the macos idiomatic location for caches. But it eliminates a code path for us, and plenty of tools use `~/.cache` on macos, among them pip, uv, and gh. 

I would say `~/.cache` is actually the more standard location for coding ecosystem tool caches, while `~/Library/Caches` is standard for application-level caches.

src/runner.rs
```diff
@@ -9,14 +9,72 @@ use std::backtrace::{Backtrace, BacktraceStatus};
...
+    if let Ok(xdg) = std::env::var("XDG_CACHE_HOME") {
+        return PathBuf::from(xdg).join("hegel");
+    }
+    let home = std::env::var("HOME").expect(
+        "Could not determine home directory: HOME is not set. \
+         Set XDG_CACHE_HOME or HEGEL_SERVER_COMMAND to work around this.",
+    );
```
Liam-DeVoe, line 37: use https://doc.rust-lang.org/std/env/fn.home_dir.html instead, which gets us windows support

src/runner.rs
```diff
@@ -266,26 +324,55 @@ fn init_panic_hook() {
...
-            return Ok(hegel_bin);
-        }
+    // Create the version directory (needed for the file lock).
+    std::fs::create_dir_all(&version_dir)
+        .map_err(|e| format!("Failed to create {}: {e}", version_dir.display()))?;
+
+    // Acquire cross-process file lock.
+    let lock_dir = version_dir.join(".install-lock");
+    acquire_file_lock(&lock_dir)?;
+    let result = do_install(&venv_dir, &version_file, &hegel_bin, &install_log);
```
Liam-DeVoe, line 345: `.install-lock` is actually a directory here, making this a lockdir, not a lockfile. I'd prefer to use a file here

src/runner.rs
DRMacIver: Sure, seems reasonable.

src/runner.rs
DRMacIver: I think Claude had some argument as to why a lockdir was better (atomic cross platform or something?) but I don't remember what it was. I'll investigate and see whether there's an actually compelling reason.

Issue #118, opened 2026-03-25
Add generators for `Duration` and `Instant`


Liam-DeVoe:
See https://doc.rust-lang.org/std/time/index.html

PR #133, opened 2026-03-26
Implement generator for `Duration`

echoumcp1:
https://github.com/hegeldev/hegel-rust/issues/118

Debating between the current implementation for `durations()` or effectively making it `integers::<u64>().map(Duration::from_nanos)` with maybe a wrapper for the `Duration`-typed bounds but no preference either way.

### DIFF
--- a/src/generators/time.rs
+++ b/src/generators/time.rs
@@ -1,6 +1,6 @@
 use super::{BasicGenerator, Generator, TestCase};
 use crate::cbor_utils::cbor_map;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 /// Generator for [`Duration`] values. Created by [`durations()`].
 ///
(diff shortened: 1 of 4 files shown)


src/generators/time.rs
```diff
@@ -0,0 +1,140 @@
...
+///
+/// #[hegel::test]
+/// fn my_test(tc: hegel::TestCase) {
+///     let i = tc.draw(hegel::generators::instants()
+///         .max_offset(Duration::from_secs(3600)));
+/// }
+/// ```
+pub fn instants() -> InstantGenerator {
+    InstantGenerator {
+        base: Instant::now(),
```
DRMacIver: I think the base being now is a bad idea, because it means you get a different instant each time the test replays, so now your tests are possibly flaky. I'd rather have a fixed base (e.g. time zero, which is presumably the unix epoch?) and let the user set it with a `base(Instant)` method.

echoumcp1: Instant is intentionally opaque and Instant::now() is the only way to construct an Instant. If the goal is to generate arbitrary points in time then I think SystemTime is what we need.

e.g.
```SystemTime::UNIX_EPOCH + std::time::Duration::from_nanoseconds(828273)```

DRMacIver: Record of conversation we had on zulip: It does indeed seem to be impossible to get a fixed Instant in rust (which is for the record insane). I suggested that we can at least fix the instant once at test start time by having a global (mutex locked) base set to Instant::now. It gets us reproducibility within a single process, and *usually* gets us reproducibility across processes unless something bad is happening.

Liam-DeVoe: this does in fact seem insane and I would like to think hard about the right thing to do here. @echoumcp1 I'd be happy to get `Duration` in in a separate PR while we decide on `Instant`

PR #135, opened 2026-03-27
Fix some test issues

DRMacIver:
Fix two flaky/broken tests:

- `TempRustProject`: PID-based temp target dirs to prevent parallel test binary races
- `test_output`: accept both `{closure#0}` and `{{closure}}` backtrace formats (Rust 1.92 changed the stable format)

### DIFF
--- a/tests/common/project.rs
+++ b/tests/common/project.rs
@@ -8,12 +8,21 @@ use std::sync::LazyLock;
 use std::sync::atomic::{AtomicU64, Ordering};
 use tempfile::TempDir;
 
-// cache build output from TempRustProject across tests. Compilation time is substantial
-// (10+ seconds) and this lets us only incur that cost on the first test.
+// Cache build output from TempRustProject across tests within the same process.
 //
-// We clear the dir at the start to ensure a fresh environment on each test run.
+// Under coverage (CARGO_LLVM_COV=1), we reuse the main project's llvm-cov target
+// dir so that subprocess builds share the same instrumented artifacts. This lets
+// cargo-llvm-cov's report phase map subprocess profraw data back to source files.
+// Cargo's internal file locks prevent corruption from concurrent access.
+//
+// Outside coverage, each test binary gets its own target dir (via PID) to avoid
+// heavy lock contention when multiple test binaries build in parallel.
 static SHARED_TARGET_DIR: LazyLock<PathBuf> = LazyLock::new(|| {
-    let path = std::env::temp_dir().join("hegel-test-cargo-target");
+    if std::env::var("CARGO_LLVM_COV").is_ok() {
+        let path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("target/llvm-cov-target");
+        return path;
+    }
+    let path = std::env::temp_dir().join(format!("hegel-test-cargo-target-{}", std::process::id()));
     let _ = std::fs::remove_dir_all(&path);
     std::fs::create_dir_all(&path).unwrap();
     path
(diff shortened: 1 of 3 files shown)


tests/common/project.rs
```diff
@@ -8,12 +8,20 @@ use std::sync::LazyLock;
...
-// (10+ seconds) and this lets us only incur that cost on the first test.
+// Cache build output from TempRustProject across tests within the same process.
 //
-// We clear the dir at the start to ensure a fresh environment on each test run.
+// Under coverage (CARGO_LLVM_COV=1), we reuse the main project's llvm-cov target
+// dir so that subprocess builds share the same instrumented artifacts. This lets
+// cargo-llvm-cov's report phase map subprocess profraw data back to source files.
+//
+// Outside coverage, each test binary gets its own target dir (via PID) to avoid
+// races when multiple test binaries run in parallel under `cargo test`.
```
DRMacIver: This seems suspicious to me. What stops us racing when run under coverage then?

RELEASE.md
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
```
DRMacIver: No release type needed for test-only changes.

tests/common/project.rs
```diff
@@ -8,12 +8,21 @@ use std::sync::LazyLock;
...
+// Outside coverage, each test binary gets its own target dir (via PID) to avoid
+// heavy lock contention when multiple test binaries build in parallel.
 static SHARED_TARGET_DIR: LazyLock<PathBuf> = LazyLock::new(|| {
-    let path = std::env::temp_dir().join("hegel-test-cargo-target");
+    if std::env::var("CARGO_LLVM_COV").is_ok() {
+        let path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("target/llvm-cov-target");
+        return path;
+    }
+    let path = std::env::temp_dir().join(format!("hegel-test-cargo-target-{}", std::process::id()));
     let _ = std::fs::remove_dir_all(&path);
```
Liam-DeVoe, line 26: Why have we changed from a shared target dir to mixing in pid outside of coverage? This lessens the benefit of a shared target dir for compilation times by a factor of `n` where `n` is the number of cores

tests/test_output.rs
```diff
@@ -44,11 +44,8 @@ fn test_failing_test_output_with_backtrace() {
...
 
-    let closure_name = if is_nightly() {
-        r"\{closure#0\}"
-    } else {
-        r"\{\{closure\}\}"
-    };
+    // Rust >= 1.92 uses {closure#0}, older stable uses {{closure}}
+    let closure_name = r"(\{closure#0\}|\{\{closure\}\})";
     // For example:
     //   Draw 1: 0
```
Liam-DeVoe, line 50: I don't think this comment about `rust >= 1.92` is correct. I'm on rust 1.93 and this test passes for me on master. I asked claude to investigate and cite this claim and it also disagreed with it, though I can't find my session about that now.

tests/common/project.rs
DRMacIver: Apparently (I haven't carefully validated this) Cargo locks the target directory, so we got ~no parallelism in the case where we are running different test processes.

tests/test_output.rs
DRMacIver: Hmm. Interesting. I'm not sure where that came from then. It emerged out of the coverage work. I'll back it out and see if we get any problems later.

tests/common/project.rs
Liam-DeVoe: I also haven't validated this but I think cargo locks during the *build* but not the test, and because the builds share the same target directory they're generally fast. When the target dir is split across processes you pay the full build cost `n` times instead of `1`, and that's where the real cost of `TempRustProject` is

PR #136, opened 2026-03-27
Fix hashset/hashmap non-basic fallback to use Collection protocol

DRMacIver:
The non-basic fallback for hashsets/hashmaps used a manual retry loop that didn't communicate duplicate rejections to the server. Now uses `Collection::new`/`more()`/`reject()`.

I originally thought this was a correctness issue but actually I think the old code was probably fine, but it's better to use the server-provided methods.


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Fix hashset and hashmap generation for non-basic element types (e.g. generators using `flat_map`). The non-basic fallback path now uses the Collection protocol (`new_collection`/`collection_more`/`collection_reject`) instead of a manual retry loop, fixing both generation correctness and shrinking quality.
```
```suggestion
RELEASE_TYPE: patch

Improve generation and shrinking of `generators::hashsets` and `generators::hashmaps`.
```

PR #137, opened 2026-03-27
Fix server crash detection

DRMacIver:
Fix the client hanging when the hegel server process exits unexpectedly:

- Drop channel senders in the background reader on pipe close, unblocking any waiting threads
- Add atexit handler to kill the server subprocess on process exit (statics aren't dropped)
- Handle server exit properly in the event loop instead of `.expect()`


src/runner.rs
```diff
@@ -121,9 +123,25 @@ impl HegelSession {
             conn_for_monitor.mark_server_exited();
         });
 
+        extern "C" fn kill_server() {
```
DRMacIver: I do not love the unsafe code here. Is this really the best way to do this?

PR #138, opened 2026-03-27
Refactor for testability

DRMacIver:
Internal refactoring in support of my 100% coverage goals. No user visible changes.

### DIFF
--- a/src/control.rs
+++ b/src/control.rs
@@ -12,6 +12,17 @@ pub(crate) fn with_test_context<R>(f: impl FnOnce() -> R) -> R {
     result
 }
 
+/// Extract a message from a panic payload.
+pub(crate) fn panic_message(payload: &Box<dyn std::any::Any + Send>) -> String {
+    if let Some(s) = payload.downcast_ref::<&str>() {
+        s.to_string()
+    } else if let Some(s) = payload.downcast_ref::<String>() {
+        s.clone()
+    } else {
+        "Unknown panic".to_string()
+    }
+}
+
 /// Returns `true` if we are currently inside a Hegel test context.
 ///
 /// This can be used to conditionally execute code that depends on a
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -221,3 +221,8 @@ pub use hegel_macros::test;
 #[doc(hidden)]
 pub use runner::hegel;
 pub use runner::{HealthCheck, Hegel, Settings, Verbosity};
+
+/// Mutex for serializing tests that modify environment variables.
+/// Used across runner, antithesis, and integration test modules to prevent races.
+#[doc(hidden)]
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -221,3 +221,8 @@ pub use hegel_macros::test;
 #[doc(hidden)]
 pub use runner::hegel;
 pub use runner::{HealthCheck, Hegel, Settings, Verbosity};
+
+/// Mutex for serializing tests that modify environment variables.
+/// Used across runner, antithesis, and integration test modules to prevent races.
+#[doc(hidden)]
+pub static ENV_TEST_MUTEX: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
Liam-DeVoe, line 228: this is a test-only mutex, why is it in a library file?

src/control.rs
```diff
@@ -12,6 +12,17 @@ pub(crate) fn with_test_context<R>(f: impl FnOnce() -> R) -> R {
...
+        s.to_string()
+    } else if let Some(s) = payload.downcast_ref::<String>() {
+        s.clone()
+    } else {
+        "Unknown panic".to_string()
+    }
+}
+
 /// Returns `true` if we are currently inside a Hegel test context.
 ///
```
Liam-DeVoe, line 27: I don't think this makes sense in `control.rs`. Why did it have to be moved out from `runner.rs`? Some requirement on what modules can import from others? If all else fails we can put this in `utils.rs`.

src/runner.rs
```diff
@@ -793,18 +806,7 @@ where
...
-                "When Hegel is run inside of Antithesis, it requires the `antithesis` feature. \
-                You can add it with {{ features = [\"antithesis\"] }}."
-            );
-
-            #[cfg(feature = "antithesis")]
-            if let Some(ref loc) = self.test_location {
-                crate::antithesis::emit_assertion(loc, !test_failed);
-            }
-        }
+        handle_antithesis_reporting(self.test_location.as_ref(), test_failed);
```
Liam-DeVoe, line 810: I think factoring this out reads strictly worse. This logic should only ever be run as part of the `Hegel::run` loop, and I'd like this logic to continue having full access to the `Hegel` context instead of passing arguments. Can we move this to a function on `Hegel` instead?

src/test_case.rs
```diff
@@ -35,15 +35,38 @@ impl std::fmt::Display for StopTestError {
...
+fn protocol_debug() -> bool {
     matches!(
         std::env::var("HEGEL_PROTOCOL_DEBUG")
             .unwrap_or_default()
             .to_lowercase()
             .as_str(),
         "1" | "true"
     )
-});
+}
```
Liam-DeVoe, line 46: Noting that this changes semantics to evaluating the environment variable at test runtime, not process startup. I don't particularly have a preference for either.

Liam-DeVoe (commented):
looks good otherwise

PR #142, opened 2026-03-27
Add script and github workflow to ratchet coverage to 100%

DRMacIver:
Claude loves avoiding coverage, and it snuck in a `--fail-under-lines 30` for what should have been a `--fail-under-lines 100`. This is partly because coverage in rust is sortof weird and broken and insisting on 100 is actually a pain in the ass for forgiveable reasons. I've brought in a custom coverage script I have from another project which parses coverage output and adds exclusion rules, and adapted it to the local context.

There's a lot of missing coverage right now. I tried to get Claude to fix this with a whole bunch of new tests but unfortunately the result was in the end mostly slop, and was better thrown away than used. Sorry for how much I made you look at of it before conceding this point, Liam.

So this PR instead adopts the traditional solution to this problem: A ratchet.

We allow nocov annotations (I am not happy about this), but track the total number of lines excluded by nocov, and the build checks that we are under that amount, and fails if we're not. Humans are allowed to raise this number if necessary, but Claude has been told that it is not (and may or may not obey this. We should keep an eye out).

Ideally (according to me) we would get this number down to zero over time, but the most important thing is to keep the amount of untested code under control, even if we never get it all the way to 0.

This also adds a lint check that coverage annotations are used in a sensible way, because claude got super weird about it and put in all sorts of slightly bizarre patterns.

Reviewing notes:

The interesting parts of this is all in the logic really. I'm much more interested in whether you're happy with the behaviour than the code.

The rust code should have no changes other than a bunch of coverage annotations (I deliberately decided to not add any tests in this PR, after previous attempts to add tests went awry).

The Python scripts are janky crap. I've read them through and they fall short of slop, but not by much. I also don't care that much because they're completely non-production-facing and are just build orchestration. If you've got strong feelings about any of them, please feel free to object.

### DIFF
--- a/scripts/check-nocov-style.py
+++ b/scripts/check-nocov-style.py
@@ -8,6 +8,7 @@
    comment (use inline // nocov instead).
 4. No // nocov end immediately followed by // nocov start (merge the blocks).
 5. // nocov start and // nocov end must be on their own lines (no code).
+6. No inline // nocov inside a // nocov start/end block (redundant).
 """
 
 from __future__ import annotations
@@ -108,6 +109,11 @@ def check() -> int:
                     # cargo fmt if you add a trailing comment.
                     stripped_end = line.rstrip()
                     block_single_line_could_be_inline = stripped_end.endswith((";", ","))
+                    # Check: no inline nocov inside a block
+                    if is_inline_nocov(line):
+                        violations.append(
+                            f"  {rs_file}:{lineno}: inline // nocov inside a // nocov start/end block (redundant)"
+                        )
                     continue
 
                 if is_inline_nocov(line):
--- a/src/generators/collections.rs
+++ b/src/generators/collections.rs
@@ -134,7 +134,7 @@ where
             let mut collection = Collection::new(tc, "composite_set", self.min_size, self.max_size);
             let mut set = HashSet::new();
             while collection.more() {
-                let element = self.elements.do_draw(tc); // nocov
+                let element = self.elements.do_draw(tc);
                 if !set.insert(element) {
                     collection.reject(Some("duplicate element"));
                     // nocov end
@@ -225,10 +225,10 @@ where
             let mut collection = Collection::new(tc, "composite_map", self.min_size, self.max_size);
             let mut map = HashMap::new();
             while collection.more() {
-                let key = self.keys.do_draw(tc); // nocov
+                let key = self.keys.do_draw(tc);
                 match map.entry(key) {
                     std::collections::hash_map::Entry::Occupied(_) => {
-                        collection.reject(Some("duplicate key")); // nocov
+                        collection.reject(Some("duplicate key"));
                     }
                     std::collections::hash_map::Entry::Vacant(entry) => {
                         let value = self.values.do_draw(tc);
@@ -327,8 +327,8 @@ impl<T: serde::Serialize, G: Generator<T>> Generator<Value> for MappedToValue<T,
 
     // nocov start
     fn as_basic(&self) -> Option<BasicGenerator<'_, Value>> {
-        let inner_basic = self.inner.as_basic()?; // nocov
-        let schema = inner_basic.schema().clone(); // nocov
+        let inner_basic = self.inner.as_basic()?;
+        let schema = inner_basic.schema().clone();
         Some(BasicGenerator::new(schema, move |raw| {
             let t_val = inner_basic.parse_raw(raw);
             crate::cbor_utils::cbor_serialize(&t_val)
--- a/src/generators/value.rs
+++ b/src/generators/value.rs
@@ -53,7 +53,7 @@ impl From<ciborium::Value> for HegelValue {
             }
             // nocov start
             ciborium::Value::Map(map) => HegelValue::Object(
-                map.into_iter() // nocov
+                map.into_iter()
                     .map(|(k, v)| {
                         let key = match k {
                             ciborium::Value::Text(s) => s,


.github/workflows/ci.yml
```diff
@@ -157,11 +157,12 @@ jobs:
     name: coverage
     runs-on: ubuntu-latest
     permissions:
-      contents: read
+      contents: write  # needed for ratchet auto-update
+      pull-requests: write  # needed for creating PRs on main
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 # zizmor: ignore[artipacked]
```
DRMacIver, line 164: What's up with this ignore?

.github/workflows/ci.yml
```diff
@@ -177,6 +178,38 @@ jobs:
...
+      - name: Auto-update coverage annotations and ratchet
+        if: success()
+        run: |
+          if git diff --quiet; then
+            echo "No coverage annotation changes needed"
+            exit 0
+          fi
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
DRMacIver: What's with this number here? That looks weird.

src/runner.rs
```diff
@@ -850,11 +882,13 @@ fn run_test_case<F: FnMut(TestCase)>(
...
                         (
-                            "<unknown>".to_string(),
-                            "?".to_string(),
-                            "<unknown>".to_string(),
-                            Backtrace::disabled(),
+                            // nocov end
+                            "<unknown>".to_string(), // nocov
+                            "?".to_string(),         // nocov
+                            "<unknown>".to_string(), // nocov
+                            Backtrace::disabled(),   // nocov
```
DRMacIver: This is a frankly bizarre combination of comments. You nocov start, then nocov end, then add a bunch of nocov to a whole bunch of consecutive lines.

src/runner.rs
```diff
@@ -866,14 +900,16 @@ fn run_test_case<F: FnMut(TestCase)>(
                     eprintln!("{}", msg);
 
                     if backtrace.status() == BacktraceStatus::Captured {
-                        let is_full = std::env::var("RUST_BACKTRACE")
-                            .map(|v| v == "full")
-                            .unwrap_or(false);
-                        let formatted = format_backtrace(&backtrace, is_full);
-                        eprintln!("stack backtrace:\n{}", formatted);
+                        let is_full = std::env::var("RUST_BACKTRACE") // nocov
```
DRMacIver: If we're going to have nocov start and end, lets use them. Please add a check to the coverage checker that looks for long sequences of nocov on 3 or more consecutive lines and requires they be replaced with start/end

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
```
DRMacIver: I don't think we need a release for this. Let's tag it as skip release and remove this file.

scripts/check-coverage.py
```diff
@@ -0,0 +1,785 @@
...
+# ///
+"""
+Custom Code Coverage Check Script
+=================================
+
+PURPOSE:
+This script provides a more nuanced coverage check than simple percentage thresholds.
+LLVM's coverage instrumentation counts regions which sometimes includes closing
+braces of control structures as separate coverage points. Additionally, certain
+patterns like todo!(), unreachable!(), and panic!() are expected to be uncovered.
```
DRMacIver: I'd say "allowed" to be uncovered.

scripts/check-coverage.py
```diff
@@ -0,0 +1,785 @@
+#!/usr/bin/env -S uv run --script
+# /// script
+# requires-python = ">=3.10"
+# ///
+"""
+Custom Code Coverage Check Script
+=================================
+
+PURPOSE:
+This script provides a more nuanced coverage check than simple percentage thresholds.
```
DRMacIver: "nuanced" is a weasel word here. "specific" maybe.

scripts/check-coverage.py
```diff
@@ -0,0 +1,785 @@
...
+   Lines that are just todo!() or todo!("message") represent intentionally
+   unimplemented code that will panic if called.
+
+3. unreachable!() Placeholders
+   Lines that are just unreachable!() or unreachable!("message") mark code
+   paths that should never be reached.
+
+4. panic!() Calls
+   Lines containing a panic!() invocation. These are defensive assertions
+   that should never be triggered in normal operation.
```
DRMacIver: No they're not. Many of them are perfectly common to trigger. But they're at least unusual, and tend to be either obviously tested or not worth testing.

DRMacIver: Actually, I think I've changed my mind about panic! Please remove the exclusion and explicitly add nocov to anything where the panic isn't covered.

scripts/check-coverage.py
```diff
@@ -0,0 +1,785 @@
...
+   the test framework compiles them but never runs them.
+
+7. // nocov Annotations
+   Lines marked with // nocov are manually excluded from line coverage.
+   Block exclusions with // nocov start ... // nocov end are also supported.
+   These are tracked by a ratchet mechanism -- the count can only decrease.
+
+RATCHET MECHANISM:
+------------------
+The count of // nocov annotations is tracked in .github/coverage-ratchet.json.
```
DRMacIver: This is now wrong, or should be. It's lines excluded, not annotations.

scripts/check-coverage.py
```diff
@@ -0,0 +1,785 @@
...
+    print(f"\nCoverage annotations: {nocov_count} // nocov")
+
+    # 6. Check ratchet
+    nocov_limit = read_ratchet()
+
+    if nocov_count > nocov_limit:
+        print(f"\nCoverage annotation ratchet EXCEEDED!")
+        print(f"  // nocov: {nocov_count} (limit: {nocov_limit})")
+        print()
+        print("You may not add new // nocov annotations without explicit")
```
DRMacIver: Please don't assume the reader of this is an LLM. Add this to your CLAUDE.md, not the tool.

.github/workflows/ci.yml
```diff
@@ -177,6 +179,40 @@ jobs:
       - name: Run coverage
         run: just check-coverage
 
+      - name: Auto-update coverage annotations and ratchet
+        if: success()
+        env:
+          GH_TOKEN: ${{ github.token }}
```
Liam-DeVoe, line 185: I'm confused why we need an auto-update job - shouldn't CI fail at PR-time if `coverage-ratchet.json` doesn't match the coverage job?

DRMacIver: It should, but we've seen enough things go out of sync when multiple merges happen that I wanted a fallback.

I'd be happy to remove this if we add "require branches to be up to date before merging" but that's a bit of a pain in the ass.

Liam-DeVoe: The code as written has confusing semantics to a reader. The intent of the person who wrote the code (you) was that we should be alerted if an edge case occurs that we don't expect to occur. A reasonable reader of the code instead reads this as "this is an integral part to the coverage ratchet workflow and can't be removed". There should be a comment here at least!

DRMacIver: Yeah, that's fair. I'll add a comment in.

PR #144, opened 2026-03-27
Add an `#[explicit_test_case]` feature.

DRMacIver:
Depends on #143. Review that one first.

This mirrors Hypothesis's example decorator. It lets you explicitly
provide a test case as a set of inputs to your test function.

Here's it in action. Consider the following:

```rust
const BAD_VALUE: u32 = 124352450;

#[hegel::test]
#[hegel::explicit_test_case(
    x = BAD_VALUE
)]
fn my_test(tc: hegel::TestCase){
    let x = tc.draw(hegel::generators::integers::<u32>());
    assert!(x != BAD_VALUE);
}
```

If you run this you will see:

```
let x = BAD_VALUE; // = 124352450

thread 'my_test' (15608116) panicked at examples/explicit_test_case.rs:9:5:
assertion failed: x != BAD_VALUE
```

It will not work in all cases. Currently in some of those cases it will likely produce very bad error messages. I've tried to do my best to think of the common cases for this and provide good error messages there, ideally at compile time, but there are 100% going to be some missing ones that we can improve over time.


Liam-DeVoe:
An argument against `always_try`: you *don't* always try it, if you disable `Phase.explicit`.

DRMacIver:
> An argument against always_try: you don't always try it, if you disable Phase.explicit.

Hmm. That's a pretty compelling argument.

How would you feel about `example` with a note saying "We're not super happy with this name, which was inherited from Hypothesis, and reserve the right to change it pre-1.0 if we can come up with a better one. The semantics will not change so this should just be a find and replace if we do." or words to that effect.

Liam-DeVoe:
yeah I'm happy with that. I'm also happy with keeping `explicit_test_case` with that same note.

DRMacIver:
> yeah I'm happy with that. I'm also happy with keeping explicit_test_case with that same note.

Mild preference for keeping `explicit_test_case` over `example` then.

src/explicit_test_case.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        // Only show the "// = debug" comment if the source and debug differ
+        // (ignoring whitespace).
+        let source_normalized: String = source.chars().filter(|c| !c.is_whitespace()).collect();
+        let debug_normalized: String = debug.chars().filter(|c| !c.is_whitespace()).collect();
+
+        if source_normalized == debug_normalized {
+            eprintln!("let {} = {};", name, source);
+        } else {
+            eprintln!("let {} = {}; // = {}", name, source, debug);
```
echoumcp1, line 91: I think a corresponding update to the output so that the format is a syntactically valid explicit test case could also be useful. 

e.g. instead of `Draw 1: 20828340`, output `#[hegel::explicit_test_case(x = 20828340)`. 

We'd have to change the actual printer `record_named_draw` in TestCase (which I cannot comment on since it is unchanged in this PR).

DRMacIver: Hmm. That's not a bad idea. It's already showing syntactically valid rust, but it's not quite the right format for `explicit_test_case`.

Does suggest that the name should be something we'd be happy to show in output, which argues in favour of just `test_case` as a name.

DRMacIver: One... not exactly problem with this, but thing to be aware of, is that we currently interleave with `note`.

This isn't necessarily an issue, but means we might e.g. want to comment-prefix notes.

So output of something like:

```rust
#[hegel::test]
fn my_test(tc: TestCase){
    let x = tc.draw(gs::integers());
    if(x == 42) note("The answer!");
    let question = tc.draw(gs::text());
}
```

Might look something like:

```rust
#[hegel::test_case(
    x = 42,
    // The answer!
    question = "What is 6 7?",
)
```

Which I don't hate as a format. I maybe even like it more than Hypothesis's.

Liam-DeVoe: I don't think we should over index on producing valid explicit example syntax. Adding explicit examples is a rare workflow compared to reading through the failure output. I'm not opposed to this, but we don't have any pressure to make a quick decision on this.

DRMacIver: Yeah I agree. I think we should assume this is going to be merged without changing the syntax for printing and then think further about the output.

But... Agreed it's rarer to add explicit examples than read failures, but maybe it wouldn't be if we made it easy enough? Hypothesis already has the git patch workflow after all.

DRMacIver:
@Liam-DeVoe I think this is probably good to merge now (I decided not to reimplement it on top of the new TestRunner abstraction. It would sortof work, but sufficiently awkwardly that it's not really a win). Do you want to do a final review?

DRMacIver:
@Liam-DeVoe I think this is probably good to merge now (I decided not to reimplement it on top of the new TestRunner abstraction. It would sortof work, but sufficiently awkwardly that it's not really a win). Do you want to do a final review?

Liam-DeVoe (approved):
happy to get this in and iterate on it. One thing top of my mind: I don't think we should have to require `#[explicit_test_case]` appear below `#[hegel::test]`, since any macro can consume any other macro, so ideally we lift this restriction in the future

Liam-DeVoe:
Something to think about: does this work with passing explicit values for a user-defined struct, eg `x = tc.draw(MyStruct.default_generator())` and `#[explicit_test_case(x = MyStruct(y: "whatever"))]`? Adding a test for that would be nice.

DRMacIver:
> Something to think about: does this work with passing explicit values for a user-defined struct, eg `x = tc.draw(MyStruct.default_generator())` and `#[explicit_test_case(x = MyStruct(y: "whatever"))]`? Adding a test for that would be nice.

Medium confident that it works but agreed we should definitely have a test for that.

PR #154, opened 2026-03-29
Improve derived generator ergonomics

### LINKED ISSUE #149: Better syntax for configuring `#[derive(DefaultGenerator)]` generators on enum tuple variants
Right now, our customization for enum tuple variants looks like this:

```rust
#[derive(DeriveGenerator)]
enum A {
    B(i32),
    C(i32, i32)
}

let g = gs::default::<A>().B(
    A::default_generator()
        .default_B().value(gs::just(42))
);
```

And I don't believe we have any way to configure tuple variants with multiple fields.

We should fix both. I think the ideal syntax looks like `gs::default::<A>().B(gs::just(42))` and `gs::default::<A>().C(gs::just(42), gs::just(43))`.

Liam-DeVoe:
I'll just restate the release notes here:

````
RELEASE_TYPE: minor

This release makes several changes to `#[derive(DefaultGenerator)]` ([#149](https://github.com/hegeldev/hegel-rust/issues/149)):

- Tuple variants now take field generators directly as positional arguments
- Named variants use a closure that receives the default variant generator:
- Generated method names are converted to snake_case instead of preserving PascalCase
  - If this would produce a name collision, we keep the original casing for both method names.

```rust
enum Op {
    Reset,
    ReadWrite(usize, usize),
    Configure { retries: u32, timeout: u64 },
}

// before
let g = Op::default_generator();
Op::default_generator()
    .ReadWrite(g.default_ReadWrite().value_0(gs::just(42)).value_1(gs::just(43)))
    .Configure(g.default_Configure().retries(gs::just(44)))

// after
Op::default_generator()
    .read_write(gs::just(42), gs::just(43))
    .configure(|g| g.retries(gs::just(44)))
```

Thanks to Rain for providing feedback on enum ergonomics!
````

I also cleaned up a bunch of the `enum_gen.rs` code in our macro, which was splitting out different code paths for no good reason.

Closes https://github.com/hegeldev/hegel-rust/issues/149. See also https://github.com/hegeldev/hegel-rust/issues/148.

### CODE REVIEW

**Comment by Liam-DeVoe** [comment]
>  don't think I like the rewriting to snake case. I know that having the pascal cased method names is extremely unidiomatic, but I think in this case it's basically the right thing to do.

I think the fact that cargo ships a default lint rule that method names be snake_case implies rust as an ecosystem follows this idiom more strongly than other languages. I don't want users to have to suppress warnings for normal hegel usage.; that'll annoy them.

> Because this is a type we entirely create ourselves, the only way you can trigger that (I think) is to have both FooBar and foo_bar as variants names, and frankly fuck you you're on your own if you do that.

If a user has chosen to do this, they have good reason to and have already suppressed a builtin cargo lint warning to do so. I want the way in which they're on their own to be "we did the best we could to provide you reasonably-named methods", *not* "we refused to generate any methods".

> I don't like the inconsistency between the tuple and named variants. I'd rather both use the closure approach and add support for tuple index updates to the variant generator.

Agree the inconsistency is bad. But:
- `.single_field(|g| g._0(gs::just(42))` feels worse than `.single_field(gs::just(42))`, because you're only configuring one field;
- and `.double_field(|g| g._0(gs::just(42)._1(gs::just(42))` feels worse than `.double_field(gs::just(42), gs::just(42))`, because `._0` and `._01` differ from the standard `.0` and `.1` tuple accessors due to syntax restrictions on method names.

**Review by DRMacIver — APPROVED** [review]

**Comment by DRMacIver** [comment]
> you can't write compose! inline there

Huh? Why not? Doesn't `enum_gen.named_field(|g| compose!(|tc| ...))` work?

> setting aside multiple methods

I don't think we should necessarily set aside multiple methods fwiw.

**Comment by Liam-DeVoe** [comment]
> Huh? Why not? Doesn't enum_gen.named_field(|g| compose!(|tc| ...)) work?

I was wrong, it does!

I've settled on an API I'm quite happy with. It involves a new `*_with` method for tuple variants that allows passing a full generator instead of configuring individual fields. `RELEASE.md` has the full details.

File: hegel-macros/src/enum_gen.rs [hunk 1]
```diff
@@ -1,126 +1,92 @@
...
+        .filter(|v| !matches!(v.fields, Fields::Unit))
         .collect();
 
+    // Compute method names for each data variant.
+    //
+    // 1. Snake-case the variant's identifier (preserving any `r#` raw prefix
+    //    that syn carries through in `to_string()`).
+    // 2. If the result would not be a valid Rust method-name token, append `_`.
+    // 3. If two variants now share a method name, fall back to the original
+    //    identifier for all colliding members.
```
DRMacIver, line 34: Please treat this as non-blocking, but here is what I would consider my ideal behaviour here:

* Add an attribute macro for variants that allows the user to rename the method name corresponding to that variant. e.g. `[hegel::default_generator_method(name=my_desired_method_name_)]`
* If anything goes wrong with our default pascal to snake case conversion (either (2) or (3)) emit either a compile error (my preference) or warning (I expect your preference) telling the user to add this attribute macro.

DRMacIver (approved):
I am basically fine with this as is, but I'd like to throw out another design consideration. Again, consider this non-blocking, it's something that I think we can fix later.

The question is this: Currently the methods that take lambdas allow you to return an arbitrary generator for the type. Should they object if I return a different variant?

e.g. consider `Op::default_generator().configure(|g| gs::just(Op::Reset))`

I believe this is perfectly valid under the current proposal. Should it be?

Despite asking the question, I weakly believe that the answer is yes. I can imagine some valid use cases, and I think it is harder to do by accident than deliberately and that nothing goes very wrong if you forbid it, even though it's a bit weird.

BUT, I think that if we're doing that, we shouldn't be skipping generators for unit variants.  If that is valid, I think it should *also* be OK to override the variant for a unit.

An example of where you might want this is if you have an enum representing a set of operations, and you want to disable an operation for one test, either by mapping it to another operation or passing it `gs::nothing()` (which we don't actually have, but I do think we should add as it serves a minor useful niche for things like this)

Liam-DeVoe:
Until we see a clear use case, I would like our stance to be that this is technically possible but not supported behavior. I don't think we can actually disallow this at the type level because rust treats enum variants as constructors to that enum, not as distinct types.

I don't buy this example:

> An example of where you might want this is if you have an enum representing a set of operations, and you want to disable an operation for one test, either by mapping it to another operation or passing it gs::nothing() (which we don't actually have, but I do think we should add as it serves a minor useful niche for things like this)

because if I wanted to disable generating via the `configure` constructor, then I would write `Op::default_generator().configure(|g| gs::nothing())`.

hegel-macros/src/enum_gen.rs
Liam-DeVoe: I do think we'll eventually want something like this, possible also respecting the similar field-renaming attribute in serde by default if present on the field. Definitely leaving this for later though

PR #160, opened 2026-03-30
Provide generators for `num` structs

### LINKED ISSUE #91: Provide generators for `num` structs
We should provide generators for [`num`](https://github.com/rust-num/num) structs like `Complex`, `BigInt`, `Ratio`, etc as an optional feature.

echoumcp1:
Closes https://github.com/hegeldev/hegel-rust/issues/91. Depends on https://github.com/hegeldev/hegel-core/pull/106


.github/coverage-ratchet.json
```diff
@@ -1,3 +1,3 @@
 {
-  "nocov": 521
+  "nocov": 548
```
DRMacIver: Why did the ratchet go up? I'd expect everything in this PR to be coverable, and the goal of the ratchet is to drive coverage to 100% so introducing new untested code is to be avoided wherever possible.

src/generators/num.rs
```diff
@@ -0,0 +1,302 @@
...
+///     assert!(*r.denom() > 0);
+///
+///     // Customize numerator and denominator ranges
+///     let r: Ratio<i64> = tc.draw(gs::rationals()
+///         .numerator(gs::integers::<i64>().min_value(0).max_value(100))
+///         .denominator(gs::integers::<i64>().min_value(1).max_value(10)));
+///     assert!(*r.numer() >= 0 && *r.denom() >= 1);
+/// }
+/// ```
+pub fn rationals() -> RationalGenerator<super::IntegerGenerator<i64>, super::IntegerGenerator<i64>>
```
DRMacIver: Feels weird to me and inconsistent with the rest of the API to have `rationals` mean a particular type of rational rather than the general interface to RationalGenerator.

src/generators/num.rs
```diff
@@ -0,0 +1,302 @@
...
+/// use num_bigint::BigInt;
+///
+/// #[hegel::test]
+/// fn my_test(tc: hegel::TestCase) {
+///     let n = tc.draw(hegel::generators::big_integers()
+///         .max_value(BigInt::from(1000)));
+///     assert!(n <= BigInt::from(1000));
+/// }
+/// ```
+pub fn big_integers() -> BigIntGenerator {
```
DRMacIver: I don't know how viable this is, but I think in my ideal world the way this would work would be that we wouldn't have a separate `big_integers` generator but `integers` would work with big integers. Seems like something we could implement by defining some traits, if the relevant traits don't already exist?

DRMacIver (changes requested):
I don't love the API. If this is the best we can do because of rust limitations then so be it, but I don't think it is and I'd really rather these feel more contiguous with the main API if possible.

.github/coverage-ratchet.json
echoumcp1: I bumped it too high by accident. Misread 540 as 548. Some are just uncovered closing braces. In `cbor_to_bigint` I added 3 no covs under the uninteresting panic cases.

DRMacIver: I don't think any of them should be uncovered closing braces - the script already filters those out, and should auto-ratchet the rest down.

Fair enough RE the uninteresting panic cases. I'd like to get us covering those in general but you shouldn't have to do that in this branch.

src/generators/num.rs
```diff
@@ -0,0 +1,174 @@
...
+impl<NG, DG> RationalGenerator<NG, DG> {
+    /// Set the numerator generator.
+    pub fn numerator<NG2>(self, numer_gen: NG2) -> RationalGenerator<NG2, DG> {
+        RationalGenerator {
+            numer_gen,
+            denom_gen: self.denom_gen,
+        }
+    }
+
+    /// Set the denominator generator. Must not produce zero.
```
DRMacIver: I wonder if it would be better to handle the zero case. 

I think there are a couple of principled things we could do here. Adding a filter would make this non-basic in cases where it should be basic, but we could add an assume or it might actually make sense to just replace zeroes with one.

src/generators/num.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn denominator<DG2>(self, denom_gen: DG2) -> RationalGenerator<NG, DG2> {
+        RationalGenerator {
+            numer_gen: self.numer_gen,
+            denom_gen,
+        }
+    }
+}
+
+impl<T, NG, DG> Generator<Ratio<T>> for RationalGenerator<NG, DG>
```
DRMacIver: It would be good for this to implement `as_basic`, because whenever `numer_gen` and `denom_gen` are basic, this can also be basic.

DRMacIver (approved):
LGTM. I'd ideally like the `as_basic` feature before merging, but it's not critical, and I don't need to rereview before merging.

Liam-DeVoe (changes requested):
The design of these generators accepts sub-generators: `rationals(gen1, gen2)`; `complex(gen1, gen2)`. Hypothesis doesn't do this, and instead accepts parameters controlling the domain of the generators: `complex(min_magnitude=..., max_magnitude=...)`, `fractions(min_value=..., maxv_value=...)`.

Accepting sub-generators gives Hegel substantially less control over the distribution, which we want because most users are going to pass bad generators. I think we should change to the API design of Hypothesis.

echoumcp1:
> The design of these generators accepts sub-generators: `rationals(gen1, gen2)`; `complex(gen1, gen2)`. Hypothesis doesn't do this, and instead accepts parameters controlling the domain of the generators: `complex(min_magnitude=..., max_magnitude=...)`, `fractions(min_value=..., maxv_value=...)`.
> 
> Accepting sub-generators gives Hegel substantially less control over the distribution, which we want because most users are going to pass bad generators. I think we should change to the API design of Hypothesis.

Two paths here I think. One is to add fractions() to the core and call into that. The other is to generate a denominator and generator a numerator within the range given the min/max value bounds, but we lose as_basic and of course it's two calls to the server. I'm leaning towards the former since calling into the server is really slow.

Liam-DeVoe:
I'd add fractions and complex into the protocol, @DRMacIver thoughts?

DRMacIver:
> I'd add fractions and complex into the protocol, @DRMacIver thoughts?

Seems basically reasonable to me yeah.

src/utils/cbor_utils.rs
```diff
@@ -0,0 +1,279 @@
...
+    let mut bytes = Vec::new();
+    ciborium::into_writer(value, &mut bytes).expect("CBOR serialization failed");
+    ciborium::from_reader(&bytes[..]).expect("CBOR deserialization failed")
+}
+
+pub fn read_value(r: &mut impl Read) -> io::Result<Value> {
+    let initial = read_u8(r)?;
+    decode(r, initial)
+}
+
```
Liam-DeVoe, line 116: `read_value`, `read_argument`, `read_raw_bytes`, and other functions below this point are new. It feels like they're reimplementing cbor parsing from scratch. We didn't need this before - what changed?

Cargo.toml
```diff
@@ -39,24 +39,36 @@ serde = { version = "1.0.103", features = ["derive"] }
 paste = "1.0"
 tempfile = "3.0"
 ciborium = "0.2.2"
+half = "2.7"
 crc32fast = "1.4"
```
Liam-DeVoe, line 43: my intuition is we don't need this `half` dependency - is it load bearing here?

Cargo.toml
```diff
@@ -39,24 +39,36 @@ serde = { version = "1.0.103", features = ["derive"] }
...
+num-traits = "0.2"
+num-integer = "0.1"
 insta = "1.47.2"
 ctor = "0.2"
 
 [features]
 default = []
 rand = ["dep:rand"]
 antithesis = ["dep:serde_json"]
+num = ["dep:num-bigint", "dep:num-rational", "dep:num-complex", "dep:num-traits", "dep:num-integer"]
```
Liam-DeVoe, line 75: since `num` splits these packages out, I think we should also split our features out, and keep the `num` feature as opting in to all the new sub-features.

Cargo.toml
echoumcp1: At the moment the server only outputs f64 (major type 7, additional info 27) constrained to the range of the requested float width, but I support f16 (additional info 25, requires half) and f32 (additional info 26) directly in the decoder for futureproofing. I can make those cases errors in the decoder instead.

Liam-DeVoe: let's remove `half` if it's unused - we shouldn't add a dependency if we don't need it!

src/utils/cbor_utils.rs
echoumcp1: ciborium eagerly decodes ints into i128, so when hegel-core returns an int beyond that range (totally valid for BigInt), it crashes. There isn't a way to specify custom decoders by type, so I partially reimplemented parsing from scratch and handle ints specially. We can use ciborium-ll directly since it gives header-by-header decoding, but that's another dependency.

Cargo.toml
echoumcp1: discussed irl - punt until we figure out the parser situation above

src/utils/cbor_utils.rs
echoumcp1: discussed irl - looking into building cbor parser on ciborium-ll and drop ciborium-io dependency

echoumcp1: I'm not convinced dropping the top level ciborium dependency is possible because we use ciborium::Value extensively.

Cargo.toml
echoumcp1: removed as a direct dependency in the latest commits, but still a transitive dependency through ciborium

PR #163, opened 2026-03-31
Add concept of different backends

DRMacIver:
This adds a notion of a "backend" to hegel-rust. A backend is essentially an abstraction over the hegel server - it provides everything you need for the test lifecycle, without reference to the fact that it's coming from hegel-core.

Part of why I want this is that I intend, later, to add a native backend. I think it will also allow for a cleaner implementation of #144. I also think even if we never do anything with it beyond this PR, it makes testing much easier. A lot of why there is so much uncovered code is that it's a pain in the ass to put the server in various broken states. In contrast, it's trivial to create a broken in-memory backend that does exactly the wrong thing we want.

All that being said, independently of the practical reasons to want such a thing, I think it makes the testing lifecycle *much* clearer, and I think ideally some abstraction like this would be present in all of our hegel client libraries. I knew this would be a nice cleanup, but I'm surprised how much more I like it than the preceding code.


src/backend.rs
```diff
@@ -0,0 +1,105 @@
...
+impl std::error::Error for StopTestError {}
+
+/// Backend for test case data generation.
+///
+/// Abstracts all communication with a data source (e.g. the hegel-core server)
+/// behind typed methods. Each fallible method returns `Result<T, StopTestError>`
+/// for operations that can be cut short by data exhaustion.
+///
+/// All methods take `&self` — implementations use interior mutability as needed.
+pub trait Backend {
```
DRMacIver: I think Backend is now an odd name for this and it should be called DataSource or something like that, as it's very scoped to a specific test case.

src/backend.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Backend ran out of data for this test case (StopTest)")
+    }
+}
+impl std::error::Error for StopTestError {}
+
+/// Backend for test case data generation.
+///
+/// Abstracts all communication with a data source (e.g. the hegel-core server)
+/// behind typed methods. Each fallible method returns `Result<T, StopTestError>`
```
DRMacIver: It's rarely triggered, but I believe that under some circumstances the backend can also return a falsified assumption. Please do some investigation to confirm this is the case. IIRC it is possible for this to happen with some combinations of floats. See https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L250

Possibly this requires reducing the number of bits to trigger this branch? https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies/_internal/numbers.py#L512

src/backend.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Begin a labeled span (used for composite generator structure).
+    fn start_span(&self, label: u64) -> Result<(), StopTestError>;
+
+    /// End the current span. If `discard` is true, the span's choices are discarded.
+    fn stop_span(&self, discard: bool) -> Result<(), StopTestError>;
+
+    /// Create a new collection. Returns an opaque handle.
+    fn new_collection(
+        &self,
+        name: &str,
```
DRMacIver: I believe name is optional in the protocol, so this should be an `Option<&str>`.

src/backend.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Returns true if a previous request triggered an abort (overflow/StopTest).
+    fn test_aborted(&self) -> bool;
+}
+
+/// Result of running a single test case.
+#[derive(Debug)]
+pub enum TestCaseResult {
+    /// Test case passed normally.
+    Valid,
+    /// Test case was rejected (assumption failed or data exhaustion).
```
DRMacIver: These should not be the same variant! They're importantly distinct cases. Probably this is a pre-existing error, but please fix it here.

src/test_case.rs
```diff
@@ -381,81 +260,60 @@ pub fn deserialize_value<T: serde::de::DeserializeOwned>(raw: Value) -> T {
...
+    handle: Option<String>,
     finished: bool,
 }
 
 impl<'a> Collection<'a> {
-    /// Create a new server-managed collection.
-    pub fn new(tc: &'a TestCase, min_size: usize, max_size: Option<usize>) -> Self {
+    /// Create a new backend-managed collection.
+    pub fn new(tc: &'a TestCase, name: &str, min_size: usize, max_size: Option<usize>) -> Self {
         Collection {
```
Liam-DeVoe, line 363: This PR reintroduces `name` for collections - but I thought we agreed collections don't need a name?

Liam-DeVoe (commented):
Thoughts on calling this notion something other than "backends"? It's an overloaded term with hypothesis backends, and also personally doesn't quite accurately communicate what it is. It makes me think we're swapping out the server implementation for another. (which we kind of are, but not really. This is actually an abstraction over the choice to use a server itself). 

This PR is the moral equivalent of adding the ability to pass different values to `.for_buffer` (now `.for_choices`), which I wouldn't really call a backend or a server. 

How about calling this a "data source", like we have in the code already, instead of a "backend"?

---

I haven't read the code super closely but I'm on board with this premise overall

src/test_case.rs
DRMacIver: Yes sorry this PR is from before we agreed that and hasn't been updated to reflect that decision.

DRMacIver:
> Thoughts on calling this notion something other than "backends"? It's an overloaded term with hypothesis backends, and also personally doesn't quite accurately communicate what it is. It makes me think we're swapping out the server implementation for another. (which we kind of are, but not really. This is actually an abstraction over the choice to use a server itself).

I think it's an overloaded term with hypothesis backend but only because they're both backends!

I think backend is the right word here. I could be persuaded by another word but I don't have a good proposal.

> 
> This PR is the moral equivalent of adding the ability to pass different values to `.for_buffer` (now `.for_choices`), which I wouldn't really call a backend or a server.

No, I don't think it is. It's literally swapping out the entire test lifecycle logic. For example, we can reimplement #144 entirely as a new backend.

> How about calling this a "data source", like we have in the code already, instead of a "backend"?

No, definitely not, it has way more control over the full testing process than being merely a data source. 

I'm happy to discuss naming a bit more but I think as long as it's an internal-only name (which I intend it to be for the foreseeable) I'd rather not worry about it too much.

DRMacIver:
How would you feel about "Test runner" over "Backend"?

Liam-DeVoe:
> For example, we can reimplement https://github.com/hegeldev/hegel-rust/pull/144 entirely as a new backend.

But this is evidence for my position, because `@example` is implemented using `.for_choices` in Hypothesis, not a backend!

I basically think we're likely to already have the words in the Hypothesis engine to describe what we're doing here better than a "backend".

I think I like "Test runner" though I haven't read this PR closely enough yet to be highly confident in that

src/runner.rs
```diff
@@ -24,6 +25,231 @@ static SESSION: std::sync::OnceLock<HegelSession> = std::sync::OnceLock::new();
...
+                    panic!("Failed to communicate with Hegel: {}", e);
+                }
+                // nocov end
+            }
+        }
+    }
+}
+
+impl DataSource for ServerBackend {
+    fn generate(&self, schema: &Value) -> Result<Value, DataSourceError> {
```
Liam-DeVoe, line 129: This is part of why I find "backend" weird: the only thing ServerBackend impl's is `DataSource`, so I would reasonably expect it to be called `ServerDataSource` and the terminology to be "data source" instead of "backend". Maybe I'm missing a further thing that `ServerBackend` does beyond impl'ing `DataSource`?

DRMacIver: Sorry, turns out we're talking past eachother because I'd forgotten the specific state I'd left this PR in. Yes, this shouldn't be called Backend. It's a leftover from an earlier version in which data sources were called backends and I went through the same thought process as you going now, and I forgot to fix this one.

I thought I had decided that what's now called TestRunner was a Backend, but apparently I'd already decided that "Test runner" was a better name for it!

DRMacIver:
> But this is evidence for my position, because @example is implemented using .for_choices in Hypothesis, not a backend!

Wait, what? How can you implement `@example` using `for_choices`? It doesn't have an underlying choice sequence.

Liam-DeVoe:
Oh I was totally thinking of `@reproduce_failure`. Strike my comment!

src/runner.rs
Liam-DeVoe: Okay great. It sounds like we're in agreement then

DRMacIver: Yup, I think so.

src/lib.rs
```diff
@@ -186,6 +186,7 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
 pub(crate) mod antithesis;
+pub mod backend;
 pub(crate) mod cbor_utils;
 pub(crate) mod control;
```
Liam-DeVoe, line 191: @DRMacIver the `backend` module should be renamed to `data_source` right? or possibly `test_runner`? (I just don't want the "backend" terminology to persist now that we've moved away from it)

PR #168, opened 2026-04-01
Improve server crash resilience: log excerpts, restart, and better error messages

DRMacIver:
We've been seeing a bunch of flaky failures on CI where the server crashes and everything fails incredibly inscrutably.

This doesn't fix those. Instead it changes it so that they should fail scrutably.

- Server crash errors now include the last few log lines inline, with long runs of indented lines folded to avoid wall-of-text stack traces.
- The control channel mutex is released before error handling, preventing `PoisonError` cascades across threads.
- `find_any` re-raises real panics instead of replacing them with the "could not find" message.
- `SESSION` is now resettable: `HegelSession::get()` detects `server_has_exited()` and spawns a new server, so subsequent `hegel()` calls work normally after a crash or explicit kill.
- A `__test_kill_server()` helper and a new `tests/test_server_restart.rs` cover the restart path end-to-end.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -946,8 +946,8 @@ fn flush_log_indent_run(
 }
 
 fn server_log_excerpt() -> Option<String> {
-    let log_path = format!("{HEGEL_SERVER_DIR}/server.log");
-    let content = std::fs::read_to_string(&log_path).ok()?;
+    let log_path = SERVER_LOG_PATH.get()?;
+    let content = std::fs::read_to_string(log_path).ok()?;
     if content.trim().is_empty() {
         return None;
     }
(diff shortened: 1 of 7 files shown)


.github/coverage-ratchet.json
```diff
@@ -1,3 +1,3 @@
 {
-  "nocov": 536
+  "nocov": 537
```
DRMacIver: I gave permission for this. There's an extra nocov in a test only function that has a case that's legitimately hard to cover and I didn't feel like insisting.

tests/test_server_crash.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// A fake server script that:
+// 1. Writes a specific error message to stderr (which goes to server.log)
+// 2. Completes the version handshake
+// 3. Exits before responding to run_test (simulating a server crash)
+//
+// The binary protocol uses 20-byte packets: magic(4) + crc32(4) + channel(4) +
+// message_id(4) + payload_len(4), followed by payload bytes and a 0x0A terminator.
+const FAKE_SERVER_WITH_LOG: &str = r#"#!/usr/bin/env python3
```
DRMacIver: All these fake servers should really be in hegel-core, I'll open a pull request with them.

src/runner.rs
```diff
@@ -844,6 +859,137 @@ fn resolve_hegel_path(path: &str) -> String {
...
+        drop(guard);
+        let _ = std::process::Command::new("kill")
+            .arg(pid.to_string())
+            .status();
+        while !conn.server_has_exited() {
+            std::thread::yield_now();
+        }
+    }
+}
+
```
Liam-DeVoe, line 1002: yuck. Surely there's a better way to do this? Can we store rust's representation of a subprocess in `SESSION`, instead of `server_pid`, and then use the native rust method (which I assume exists, but have not checked) to shut down the process?

DRMacIver: Yeah, definitely. I had missed that it was calling the kill command. Yikes.

Even if we just call kill on the pid from here that would still be an improvement.

PR #173, opened 2026-04-04
Re-raise internal errors instead of swallowing

### LINKED ISSUE #47: Re-raise internal errors instead of swallowing
Right now our panic hook swallows any internal errors. We should inspect the traceback for the filename / exception type and reraise any exception that is hegel-internal.

Liam-DeVoe:

Previously, when hegel-internal code panicked during test execution (e.g. invalid generator config like `integers().min_value(100).max_value(10)`, or an unexpected server error response), the panic was caught by `catch_unwind`, treated as a test failure, and reported as "Property test failed: unknown" — completely losing the actual error. This follows the same pattern as Hypothesis's `_execute_once_for_engine`, which re-raises exceptions that originate from Hypothesis's own source files.

Now we check the panic location's file path against `CARGO_MANIFEST_DIR` to detect hegel-internal panics and propagate them immediately on the first test case — no shrinking, no wrapping. The error message includes the original source location and backtrace so users can trace where things went wrong.

</details>

Fixes #47.


DRMacIver:
I think this problem is now fixed with the use of `hegel_internal_assert`

PR #188, opened 2026-04-16
Native backend for hegel rust

DRMacIver:
WARNING: This is for preview purposes only. It's not at all ready for review.

This is the result of asking Claude to write a native backend for hegel-rust and letting it iterate on it in a test-driven manner.

It currently has some known omissions that Claude is still working on. I'm putting this up just for people to look at and to give it access to the full CI as part of its iteration loop.


DRMacIver:
Closing this to reduce build spam. Will reopen later.

PR #189, opened 2026-04-16
Add Windows support

DRMacIver:
Basic support for Windows in `hegel-rust`.

Main limitation is that the uv auto-install doesn't work, because I wasn't confident in getting that working robustly and it doesn't seem too bad to ask people to install uv on windows.

Depends on https://github.com/hegeldev/hegel-core/pull/95 and will need to be updated to remove temporarily pointing hegel-core to that branch before it is merged.


.github/workflows/ci.yml
```diff
@@ -83,12 +109,18 @@ jobs:
...
           tools: just uv
 
+      - if: runner.os == 'Windows'
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
       - name: Run tests with all features
-        run: just check-tests-all-features
+        env:
+          RUST_BACKTRACE: "1"
+        run: cargo test ${{ matrix.features }}
```
Liam-DeVoe: our ci is now out of sync with our justfile which I really want to avoid. Can we add a new justfile command `just check-tests-rand-feature` alongside `just check-tests-all-features`? Yes these names are terrible, but I prefer verbose ci names if it maintains the invariant that anything ci runs can be reproduced via lockfile commands

.github/workflows/ci.yml
```diff
@@ -59,21 +61,45 @@ jobs:
...
+      # uv on Windows. Set up uv here; once it's on PATH, hegel-rust drives
+      # hegel-core installation via uv the same way it does on Linux.
+      - if: runner.os == 'Windows'
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
       - name: Run tests
-        run: just check-tests
+        env:
+          RUST_BACKTRACE: "1"
+        run: cargo test
```
Liam-DeVoe, line 72: why did this change from `just check-tests`?

tests/embedded/runner_tests.rs
```diff
@@ -49,14 +57,23 @@ fn test_settings_new_in_ci_disables_database() {
...
 fn test_wait_for_exit_timeout() {
+    #[cfg(unix)]
     let mut child = Command::new("sleep").arg("100").spawn().unwrap();
+    #[cfg(windows)]
+    let mut child = Command::new("cmd")
+        .args(["/C", "ping -n 100 127.0.0.1 >nul"])
+        .spawn()
+        .unwrap();
     let result = wait_for_exit(&mut child, Duration::from_millis(50));
     assert!(result.is_none());
```
Liam-DeVoe, line 93: oh boy...

tests/test_antithesis.rs
```diff
@@ -1,3 +1,6 @@
+// Antithesis only supports Linux.
+#![cfg(not(windows))]
+
 mod common;
```
Liam-DeVoe, line 4: I'm somewhat confused - if `test_antithesis` is only run when the antithesis feature is enabled, and we never enable the antithesis feature on windows, then this cfg shouldn't be necessary? Or rather, how did this file work at all previously when it requires the antithesis feature?

.github/workflows/ci.yml
```diff
@@ -83,12 +109,18 @@ jobs:
...
-      - uses: ./.github/actions/install-tools
+      - if: runner.os == 'Linux'
+        uses: ./.github/actions/install-tools
         with:
           tools: just uv
 
+      - if: runner.os == 'Windows'
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
       - name: Run tests with all features
```
Liam-DeVoe: I'd prefer to push this logic down into `install-tools`, which should check the runner os and decide the appropriate install method and version. This current logic is asking for trouble when we update the uv version.

Liam-DeVoe (commented):
otherwise lgtm

tests/embedded/runner_tests.rs
DRMacIver: lolwut. Sorry.

I either completely failed to see that or it snuck in since I last reread the code.

.github/workflows/ci.yml
DRMacIver: Good question. I suspect Claude thinks just doesn't work on windows.

PR #192, opened 2026-04-20
Introduce `repeat` API primitive

DRMacIver:
This adds `tc.repeat` as a primitive, for a test-case local version of `loop` - operations that you want to run logically forever, or at least until they hit an error.


src/test_case.rs
```diff
@@ -210,10 +251,101 @@ impl TestCase {
...
+        // around float precision when min_size approaches 2^53, so we add a
+        // cap well below that (the backend's data budget runs out long before
+        // this many iterations anyway) without tripping the bug.
+        const MAX_SAFE_MIN_SIZE: usize = 1 << 40;
+        let min_size = self.draw_silent(integers::<usize>().max_value(MAX_SAFE_MIN_SIZE));
+
+        let mut collection = Collection::new(self, min_size, None);
+        let mut iteration: u64 = 0;
+
+        while collection.more() {
```
DRMacIver, line 327: Something worth noting: I intend to make this into a proper infinite loop when run in `one_shot` mode.

src/test_case.rs
```diff
@@ -210,10 +251,101 @@ impl TestCase {
...
+        let indent = local.indent;
+        (local.on_draw)(&format!("{:indent$}{}", "", message, indent = indent));
+    }
+
+    /// Run `body` in a loop that should runs "logically infinitely" or until
+    /// error. Roughly equivalent to a `loop` but with better interaction with
+    /// the test runner: This loop will never exit until the test case completes.
+    ///
+    /// At the start of each iteration a `// Loop iteration N` note is emitted
+    /// into the failing-test replay output.
```
Liam-DeVoe, line 291: I'd prefer `// Repeat iteration N`. If I see "loop" I might not immediately correlate that to `tc.repeat`, especially since rust has an actual language construct called `loop`

justfile
```diff
@@ -61,3 +61,10 @@ test: check-tests
 coverage: check-coverage
 lint: check-lint
 check: check-lint check-tests check-tests-all-features
+
+# Run cargo-insta, installing it first if it's missing. Pinned to the same
+# version as the `insta` dev-dependency so snapshot format stays consistent.
+INSTA_VERSION := "1.47.2"
+insta *ARGS:
+    @cargo install --quiet --locked --version {{INSTA_VERSION}} cargo-insta
+    cargo insta {{ARGS}}
```
Liam-DeVoe, line 70: no, justfile should not be installing dependencies. Claude needs to learn that when it adds a dependency, it needs to rerun `cargo install` in its local environment. The justfile also should not be pinning the insta version, that's the lockfile's job.

Liam-DeVoe (commented):
is `insta` actually used anywhere in this PR? I don't see any new files that would contain the saved snapshot contents

Liam-DeVoe (approved):
(api and names LGTM)

DRMacIver:
> is `insta` actually used anywhere in this PR? I don't see any new files that would contain the saved snapshot contents

Insta asserts are updated inline in the source code.

Liam-DeVoe:
oh yes I see that now:

```
    insta::assert_snapshot!(output, @r"
    // Loop iteration 1
      let x_1 = 10;
    ");
```

I like this ux better than syrupy tbh I think (I have put all of 30 seconds of thought into it)

justfile
DRMacIver: Irksomely, that doesn't work. 

Cargo commands aren't installable via your Cargo.toml, because they're external binaries named e.g. `cargo-insta` rather than dev dependencies.

I think the UX of having to manually type in the install yourself is much worse than installing in the justfile.

src/test_case.rs
DRMacIver: Yup, makes sense. Leftover from when I originally called it `loop` before I remembered that there was a language construct of the same name.

DRMacIver:
> I like this ux better than syrupy tbh I think (I have put all of 30 seconds of thought into it)

I think there are significant ux trade-offs each way, and this style tends to work better for small snapshots and external files tend to work better for large ones.

justfile
Liam-DeVoe: goddamnit. OK, looks good for now

PR #193, opened 2026-04-20
Add #[hegel::main] and #[hegel::standalone_function] entry points

DRMacIver:
This adds two new macros to allow more flexible use of hegel. Closes #153 

- #[hegel::main] wraps a function as a standalone binary entry point, exposing CLI flags for every Settings option.
- #[hegel::standalone_function] rewrites fn(tc: TestCase, args...) into fn(args...) so a property test can be invoked directly.

`hegel::main` is the primary goal of this, because it makes it easier to use hegel in workload contexts, but in #153 we'd talked about also providing a way of invoking hegel tests directly, so it seemed worth adding that at the same time.

I am very much not in love with the name `standalone_function` and would appreciate suggestions of better names, but I couldn't come up with any off the top of my head that weren't worse. Consider it a placeholder in the spirit of `explicit_test_case`

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This adds two new macros to allow more flexible use of hegel.
+
+* `#[hegel::main]` wraps a function as a standalone binary entry point, exposing CLI flags for every Settings option.
+* `#[hegel::standalone_function]` rewrites fn(tc: TestCase, args...) into fn(args...) so a property test can be invoked directly.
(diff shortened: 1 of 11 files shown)


Liam-DeVoe (approved):
lgtm. (this is just a review of the api; only a minimal skim of the code).

A weak naming suggestion: `hegel::property`. I don't love it because it's overloading a general term to mean something a bit too specific, but I think it's kind of reasonable in this case.

DRMacIver:
> lgtm. (this is just a review of the api; only a minimal skim of the code).
> 
> A weak naming suggestion: `hegel::property`. I don't love it because it's overloading a general term to mean something a bit too specific, but I think it's kind of reasonable in this case.



> lgtm. (this is just a review of the api; only a minimal skim of the code).
> 
> A weak naming suggestion: `hegel::property`. I don't love it because it's overloading a general term to mean something a bit too specific, but I think it's kind of reasonable in this case.

I think I like this less than `standalone_function` because it suggests to me that it would run as a test and I anticipate a lot of confused users wondering why it doesn't.

PR #194, opened 2026-04-20
Add extremely fragile support for threading

DRMacIver:
This adds the correct API for threading (TestCases can be moved between threads but not shared. You can clone one to pass it to another thread) with the worst implementation - just put a lock around everything.

This is an intentionally bad implementation in order to unblock some important use cases around workloads, with a lot of "here be dragons" warnings around it to make sure people realise how bad it is.

My expectation is that in the short term the only real way to use it safely will be in one-shot mode (and even that is not 100% safe).

Note: Currently `explicit_test_case` is not supported for tests that try to generate data from multiple threads. This is fixable but it doesn't currently seem worth it, as the cases where you want each don't overlap much.


Cargo.toml
```diff
@@ -40,6 +40,7 @@ paste = "1.0"
 tempfile = "3.0"
 ciborium = "0.2.2"
 crc32fast = "1.4"
+parking_lot = "0.12"
 hegeltest-macros = { version = "=0.6.2", path = "hegel-macros" }
```
Liam-DeVoe: I generally avoid new dependencies more than you do, but I'm not qualified to judge this particular one, so I'm trusting your judgment here!

Liam-DeVoe (approved):
I am not qualified to review this, so treat this review as "nothing here looks insane" rather than "I have actually thought about this design"

Cargo.toml
DRMacIver: Hmm. I'm not sure about this one either actually. Let me investigate but I think it's probably unnecessary.

DRMacIver: On further review, I think it's legit. The big thing it gives us is that rust standard library mutexes get poisoned when you panic from within claiming them, and parking lot ones don't. I think it's also pretty uncontroversial as dependencies go - widely used, fairly stable.

Liam-DeVoe: yeah seems like a pretty common dependency, so I'm not opposed

DRMacIver: TBH I think in general we just have to be more comfortable with dependencies in rust, because it's a minimalist rather than maximalist standard library.

Still worth thinking about each one, but the answer is much more likely to be that it's worth it than in Python.

PR #197, opened 2026-04-20
Improvements to TempRustProject performance and isolation

DRMacIver:
I was running into some fairly brutal performance problems with repeatedly running tests in #188, and on top of that when working on many PRs simultaneously the shared directory in /tmp would often corrupt builds leading to flakiness.

This PR should solve both problems. I have not ultra-carefully vetted it (especially the comments) but it seems mostly sane and appears to work.

What Claude says: 

<details>
<summary>Summary and benchmarks</summary>

## What changed

Two test-infrastructure changes, both in `tests/common/project.rs` (plus a regex update in `tests/test_output.rs` to match the new crate-name shape in backtraces).

### 1. Move the shared cargo target dir out of `/tmp`

The `TempRustProject` cargo target cache lived at `/tmp/hegel-test-cargo-target`. Every parallel checkout of the repo shared that path, and each `cargo test` run wiped it on the first `LazyLock` touch — so running tests in one checkout clobbered the warm caches of every other checkout. The wipe also meant that back-to-back `cargo test` runs in the same checkout both paid the cold-build cost.

Moved to `CARGO_TARGET_TMPDIR/hegel-shared-target`, which is under the workspace's own `target/`. Each checkout now has its own cache, the cache persists across runs, and cleanup is delegated to `cargo clean`.

### 2. Warm up the shared target once per process instead of racing

A cold `cargo build` of `hegeltest` and its transitive deps peaks at hundreds of MB of rustc memory. Without any coordination, the first wave of parallel test threads would each kick off their own cold build and risk OOM-killing rustc (in addition to wasting CPU on duplicate startup). Cargo's `.cargo-lock` already serialises the compile phase, but it doesn't prevent N processes from each doing their own cold build.

Added a `OnceLock` gate: the first thread to need a subprocess cargo invocation runs a dedicated warmup `cargo build` against a synthetic crate that depends on `hegeltest`, while every other thread blocks. After warmup the target dir is hot, and subsequent per-test cargo calls (mostly linking a tiny wrapper crate) can all run concurrently without contending for build work.

### 3. Hardening for shared-target parallelism

Two flakiness risks surfaced on review against concurrent test-binary processes (which `cargo test` spawns in parallel across integration tests):

- **Crate-name collision.** `temp_hegel_test_N` used a process-local `AtomicU64`, so every test-binary process started the counter at 0 and two of them would both try to produce `target/debug/temp_hegel_test_0` (cargo's shallow output copy, which is not hash-suffixed). With the old `/tmp` target this was also a latent bug but evidently rare; the new setup makes it more plausible. Fixed by prefixing the name with the PID so it's globally unique across siblings.
- **Warmup truncate race.** The warmup `Cargo.toml` / `src/lib.rs` writes used plain `fs::write`, which `O_TRUNC`s in place. A peer process's cargo could briefly observe a truncated file. Fixed by writing to a PID+counter-suffixed temp path and atomically renaming, so readers see old-or-new, never partial.

## Benchmarks

Same host, 13 TempRustProject-heavy integration-test binaries (including the mixed-feature `test_antithesis` case) plus full `cargo test`:

| Scenario | Before (main) | After | Speedup |
|---|---|---|---|
| 13 TempRustProject binaries, cold shared target | 108 s | 31 s | **3.5×** |
| 13 TempRustProject binaries, re-run | 86 s¹ | 24 s | **3.6×** |
| Full `cargo test` (36 binaries) | 113 s | 61 s | **1.9×** |

¹ The before-re-run is still ~cold because the old code wiped `/tmp/hegel-test-cargo-target` on every first `LazyLock` touch — there was no "warm" state to return to.

The "only 1.9×" on the full suite is because non-TempRustProject tests (shrink-quality, hegel-server-backed tests, etc.) dominate the 61 s steady-state; those are unaffected. Isolating just the tests that actually spawn cargo subprocesses, the speedup is the 3.5–3.6× figure.

## Stress-testing

Adversarially stress-tested the new path: 5 consecutive cold runs of all 13 TempRustProject-using integration-test binaries in parallel, covering the mixed-feature case where cargo must keep both `antithesis`-flavored and default-flavored artifact variants side by side in the same target. 13/13 pass every iteration.

## What I deliberately did *not* port from #188

The original changes were part of the native-backend preview PR; most of the rest isn't applicable on `main`:

- The `Invocation<'_>` API refactor (reuses one built project across multiple `#[test]`s) — not a perf change and would touch every caller without benefit here.
- `features = ["native"]` default on `cfg!(feature = "native")` — feature doesn't exist on main.
- `__CARGO_LLVM_COV_RUSTC_WRAPPER_CRATE_NAMES` coverage plumbing — current coverage CI already works.
- `HealthCheck::TooSlow` / `FilterTooMuch` suppressions in `tests/common/utils.rs` — motivated by native-backend speed quirks.
- The trybuild migration of compile-fail tests — I measured this and with the shared-target + warmup changes in place, the marginal saving on compile-fail tests is only ~1–2 s warm (each subprocess rustc emits the expected error in ~30 ms once the target is hot). Left out to keep this PR focused on the big wins; happy to do it separately if desired.

</details>

### DIFF
--- a/tests/common/project.rs
+++ b/tests/common/project.rs
@@ -27,7 +27,12 @@ fn shared_target_dir() -> PathBuf {
 }
 
 // use a unique package name in our Cargo.toml to avoid lock contention of parallel
-// cargo builds within the shared target dir.
+// cargo builds within the shared target dir. The PID prefix disambiguates across
+// sibling test-binary processes that all share `CARGO_TARGET_DIR`: without it,
+// each process starts the counter at 0 and they'd collide on
+// `target/debug/temp_hegel_test_0` (the "shallow" output copy cargo produces for
+// binary crates), which can fail with ETXTBSY if another process's copy is still
+// running.
 static PACKAGE_NAME_ID: AtomicU64 = AtomicU64::new(0);
 
 // First-build gate. A cold `cargo build` of `hegeltest` and its transitive
@@ -58,6 +63,18 @@ fn warmup_shared_target() {
     });
 }
 
+fn write_atomic(path: &Path, content: &[u8]) {
+    // PID + counter keeps the temp name unique across concurrent processes
+    // and across concurrent calls in the same process. `rename(2)` is atomic
+    // on POSIX: a concurrent reader of `path` never sees an in-progress write,
+    // only the old or new full content.
+    static TMP_COUNTER: AtomicU64 = AtomicU64::new(0);
+    let tmp_id = TMP_COUNTER.fetch_add(1, Ordering::Relaxed);
+    let tmp = path.with_extension(format!("tmp.{}.{}", std::process::id(), tmp_id));
+    std::fs::write(&tmp, content).unwrap();
+    std::fs::rename(&tmp, path).unwrap();
+}
+
 fn run_warmup_build(hegel_path: &Path, shared_target: &Path) {
     let warmup_dir = shared_target.join("warmup");
     std::fs::create_dir_all(warmup_dir.join("src")).unwrap();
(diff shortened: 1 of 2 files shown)


Liam-DeVoe:
I don't love the warmup logic existing, otherwise looks fine on a skim

DRMacIver:
> I don't love the warmup logic existing, otherwise looks fine on a skim

I think it makes sense. It's a pretty big speed up on compile times in tests, and I was finding those very painful, and while I don't love how it works I do love that we can pay that cost only once.

PR #198, opened 2026-04-20
Add TestCase::reject() with never-return type

DRMacIver:
I know you prefer `assume(false)`, @Liam-DeVoe, but a) I'm fond of `reject()` but b) most importantly, in rust in particular this has a static type that lets the compiler realise that code after it is unreachable and allows us to use it in arbitrary expressions, which I think is actually super important ergonomics.


Liam-DeVoe:
lgtm. registering a desire to at least have a conversation before we port this to other languages without the same static benefit though

PR #201, opened 2026-04-21
Add single test-case mode to rust client

DRMacIver:
Client-side changes corresponding to https://github.com/hegeldev/hegel-core/pull/97.

NB: PR will need updating once that one's been merged.

<details>
<summary>Details</summary>

- Adds `Mode` enum (`TestRun` / `SingleTestCase`) and `Settings::mode()` builder method
- Adds `#[hegel::main]` attribute macro that wraps a test function in `fn main()` with `--single-test-case` CLI flag
- CI temporarily installs hegel-core from the `DRMacIver/one-shot` branch (marked with TODOs to revert once #97 is released)

</details>


DRMacIver:
I think the failures are all because of the hack to use hegel-core from the right branch. Once that's fixed they should go away, or I'll otherwise fix them.

tests/test_single_test_case.rs
```diff
@@ -0,0 +1,219 @@
...
+
+#[test]
+fn test_single_test_case_assume_produces_invalid() {
+    hegel::Hegel::new(|tc| {
+        let n = tc.draw(gs::integers::<i32>());
+        tc.assume(n > 0);
+    })
+    .settings(hegel::Settings::new().mode(Mode::SingleTestCase))
+    .run();
+}
```
Liam-DeVoe, line 93: right, we'll have to figure out what to do about retrying failed assumptions in this mode...

Liam-DeVoe (approved):

tests/test_single_test_case.rs
DRMacIver: I think at the top level we don't retry failed assumptions. I think `hegel::main` should maybe have a special exit code it uses to indicated it failed with a top-level falsified assumptions. Maybe `hegel::test` should actually panic.

DRMacIver: An alternative is to wrap the whole top level in an implicit

```python
while True:
    try:
         ...
    except UnsatisfiedAssumption:
        pass
```

DRMacIver: But I think that has more problems than it solves and if you want that you should put your test in a tc.repeat

Liam-DeVoe: yeah I'm genuinely not sure, we should talk about this synchronously sometime (this is not a blocker for this PR)

PR #204, opened 2026-04-22
Add deferred initialisation generators

DRMacIver:
When defining mutually recursive generators you sometimes need to forward reference. This API lets you declare generators before you define them in order to achieve that.

<details>
<summary>Implementation details</summary>

`DeferredGenerator<T>` wraps an `Arc<OnceLock<BoxedGenerator<'static, T>>>`. It is `Clone` (via `Arc` clone) and implements `Generator<T>` by delegating to the inner generator.

- `gs::deferred::<T>()` creates an empty slot
- `.set(generator.boxed())` fills it exactly once (panics on double-set)
- `.do_draw()` delegates to the inner generator (panics if not yet set)
- No `as_basic()` — deferred generators always use span-based generation

### Example: mutual recursion

```rust
let x = gs::deferred::<i32>();
let y = gs::deferred::<i32>();

y.set(hegel::one_of!(gs::integers::<i32>().min_value(0).max_value(10), x.clone()).boxed());
x.set(hegel::one_of!(gs::integers::<i32>().min_value(100).max_value(110), y.clone()).boxed());
```

### Tests

- Basic delegation, clone sharing, composition with `map`
- Panic on double-set and draw-before-set
- Self-recursive generator (growing a `Vec<bool>`)
- Mutually recursive generators reaching values through both paths

</details>


.claude/CLAUDE.md
```diff
@@ -106,3 +106,7 @@ This project enforces 100% line coverage for new code. You may not add `// nocov
 ### Conformance Tests
 
 Located in `tests/conformance/`. Rust test binaries in `tests/conformance/rust/src/bin/` are invoked by a Python test runner (`tests/conformance/test_conformance.py`) that validates generators produce values matching their declared constraints.
+
+### Pull Requests
+
+When creating a pull request, always create it as a **draft**. The human must provide the title and a short description. Any AI-generated content (implementation details, test summaries, etc.) must go inside a `<details>` block after the human-provided description.
```
Liam-DeVoe: I think this is you-specific - I have conflicting instructions to my agents! 

Put this in your CLAUDE.local.md instead?

DRMacIver: Happy to pull this from the PR for now, but I'm keen for us to find something we can put in here, because I want something that tells external contributor's claudes how to do this right.

Liam-DeVoe: let's talk irl

PR #206, opened 2026-04-23
feat: add PathBuf DefaultGenerator

andrewgazelka:
## Summary
- add a `DefaultGenerator` impl for `PathBuf`
- generate common filesystem edge cases including traversal, empty segments, hidden names, long names, whitespace, and occasional absolute paths
- keep `#[derive(DefaultGenerator)]` working for structs that contain `PathBuf` fields

## Why
Downstreams using Hegel for filesystem models need structured path generation without writing a custom generator for every type that embeds a `PathBuf`.

### DIFF
--- a/src/generators/default.rs
+++ b/src/generators/default.rs
@@ -6,6 +6,7 @@ use super::{
 };
 use std::collections::HashMap;
 use std::hash::Hash;
+use std::path::PathBuf;
 use std::time::Duration;
 
 /// Trait for types that have a default generator.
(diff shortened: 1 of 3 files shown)


andrewgazelka:
unsure about this.... Path is OS dependent. thoughts?

andrewgazelka:
tbh i think this AI code is shit. however, I do not really know what it should be

src/generators/default.rs
```diff
@@ -6,8 +6,48 @@ use super::{
...
+    use super::{just, one_of};
+
+    one_of([
+        text().max_size(12).boxed(),
+        just(".".to_string()).boxed(),
+        just("..".to_string()).boxed(),
+        just(String::new()).boxed(),
+        text()
+            .min_size(1)
+            .max_size(8)
```
andrewgazelka, line 22: seems constraining

src/generators/default.rs
```diff
@@ -6,8 +6,48 @@ use super::{
...
+        return path;
+    }
+
+    absolute_path_root().join(path)
+}
+
+fn absolute_path_root() -> PathBuf {
+    #[cfg(windows)]
+    {
+        PathBuf::from("C:\\")
```
andrewgazelka, line 43: this is where pathbuf seems like it might have issues. is it supposed to be an existing path for instance or no? should users just impl generator themselves? sucks if you have a strcut that arleady uses a Path but... oh well

Liam-DeVoe:
Hypothesis has the following default generator: 

```
os.PathLike: st.builds(PurePath, st.text())
```

which seems fine to me to start! 

We don't actually have a `paths()` generator in Hypothesis, so there isn't prior art here. Getting a good generator is probably subtle enough that we want it as part of the protocol rather than reimplemented per-library.

src/generators/default.rs
Liam-DeVoe: (I think we should take a simpler approach entirely, see my comment below, but to answer the question as-is here:) We shouldn't constrain ourselves to only generating existing paths by default. We might imagine a `exists_only: bool = False` parameter in the supposed `paths()` generator, but getting the performance of that right is nontrivial, and probably also needs to come with a security warning.

src/generators/default.rs
Liam-DeVoe: yep this should be completely unconstrained. the models love producing unjustifiably-constrained generator code

Liam-DeVoe:
PathBuf now implements DefaultGenerator: https://github.com/hegeldev/hegel-rust/pull/215

PR #207, opened 2026-04-23
Fix ledger example clippy map iteration

andrewgazelka:
## Why this is needed
- `cargo clippy --all-targets --all-features -- -D warnings` currently fails on `examples/ledger.rs` with `clippy::for_kv_map`
- that blocks the repo's warning-free clippy gate for a no-behavior-change issue in the example

## Summary
- replace key/value iteration with `HashMap::values()` in the ledger invariant
- keep the example behavior unchanged while satisfying `clippy::for_kv_map`

## Validation
- `cargo clippy --all-targets --all-features -- -D warnings`

### DIFF
--- a/examples/ledger.rs
+++ b/examples/ledger.rs
@@ -77,7 +77,7 @@ impl LedgerTest {
 
     #[invariant]
     fn nonnegative_balances(&mut self, _: TestCase) {
-        for (_account, balance) in &self.ledger.balances {
+        for balance in self.ledger.balances.values() {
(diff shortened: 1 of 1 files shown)


Liam-DeVoe:
I'm surprised this wasnt caught in our CI, do we not run clippy over the examples? We should fix that!

DRMacIver:
Thanks!

And yes, agreed we should fix our CI to check clippy on the examples.

PR #208, opened 2026-04-23
feat: add a public custom runner hook

andrewgazelka:
## Summary
- add a public `Hegel::run_with_runner(...)` hook for alternative backends
- expose read-only `Settings` accessors needed by external runners
- add a regression test covering a custom runner implementation

## Why
This lets downstreams plug in non-server execution backends without forking Hegel's test builder and panic/shrinking wiring.

### DIFF
--- a/.github/actions/install-tools/action.yml
+++ b/.github/actions/install-tools/action.yml
@@ -16,11 +16,9 @@ runs:
           esac
         done
       shell: bash
-    - if: steps.parse.outputs.just == 'true' && runner.os == 'Linux'
-      run: sudo apt-get update && sudo apt-get install -y just
-      shell: bash
-    - if: steps.parse.outputs.just == 'true' && runner.os == 'Windows'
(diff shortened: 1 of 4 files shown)


DRMacIver:
Not in principle opposed, but could you talk more about your use case? Right now the backends feature is somewhat internal only and we're not necessarily committing to it looking like that long-term (though it probably will), so I'd like to understand what you're trying to do before adding new API surface for it.

andrewgazelka:
> Not in principle opposed, but could you talk more about your use case? Right now the backends feature is somewhat internal only and we're not necessarily committing to it looking like that long-term (though it probably will), so I'd like to understand what you're trying to do before adding new API surface for it.

I was trying to add fuzzing support to hegel, but I realized I was doing it the wrong way (this replaces hypothesis logic afaik). Instead I want to do it with `tc.target` with a number derived from code coverage via `sancov`. playing around with that internally rn.

this can be closed

PR #212, opened 2026-04-24
Avoid sharing .hegel directory between subsequent runs of cargo test

DRMacIver:
<details>

## Summary
- Library behaviour is unchanged — `.hegel/` still lives in the caller's cwd for real users.
- Every integration-test binary in this repo now chdir's into a fresh tempdir (`/tmp/hegel-rust-test-*`) before `main` runs, via a `#[ctor::ctor]` in `tests/common/mod.rs`. The tempdir is held in a `OnceLock` for the process lifetime.
- The 11 test files that didn't already pull in `common` get a bare `mod common;` declaration so the ctor is linked in.
- Added `ctor = \"0.2\"` as a dev-dependency.

## Test plan
- [x] New `tests/test_runtime_dir_isolation.rs` asserts that cwd is no longer the crate root, that it starts with `hegel-rust-test-`, and that running a hegel test creates `.hegel/` inside the tempdir (not in the crate root).
- [x] `just check` is green (88 test-result blocks, 0 failures).
- [x] `just check-coverage` is green (0 uncovered new lines).

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,6 +103,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
+name = "ctor"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


Liam-DeVoe:
what's the actual problem this is solving? The hypothesis test suite doesn't do this, for example - if a Hypothesis test outcome depends on the local .hypothesis directory state, then we typically set `@settings(database=None)` for that test.

PR #214, opened 2026-04-25
Add `generate_with` annotation

ChernoffBound:
Added the ability to annotate with #[hegel(generate_with = "path")] on fields in a struct using #[derive(DefaultGenerator)] to specify custom function for generating the field.
This gives the user the ability to easily incorporate support for unsupported types such as chrono and serde::Value


DRMacIver:
Thanks! This is a useful feature.

Question: I'm not super sure about some of the specific design decisions. It strikes me as slightly odd that the argument is called "path" and that it's passed as a string rather than a rust expression. Is there a particular reason behind those choices? (I confess I'm not actually very au fait with rust macro API designs so it's totally plausible I'm missing something and this is a normal thing to do)

ChernoffBound:
I have modeled the semantics off of serde's serialize_with annotation: https://serde.rs/field-attrs.html#serialize_with.
Due to how the syn code parses the string into a syn::Path in hegel-macros/src/struct_gen.rs:49, a non-existent path would still give you an error saying that the path does not point to a valid function

Liam-DeVoe:
I think it's possible we'll eventually want something like this, but I'd like to hold off for now and see how far we can get with improving the ergonomics of configuring the default generator instead.

PR #215, opened 2026-04-25
claude: add DefaultGenerator impl for PathBuf

Liam-DeVoe:

Adds a `DefaultGenerator` impl for `PathBuf`, so `#[derive(Generate)]` works for structs with `PathBuf` fields and `gs::default::<PathBuf>()` returns a generator.

Implementation mirrors Hypothesis's default for `os.PathLike` (`st.builds(PurePath, st.text())`): `text().map(PathBuf::from).boxed()`. As discussed in #206, getting a richer path generator right is subtle and probably belongs at the protocol level (a shared `paths()` strategy) rather than reimplemented per-library — so this just unblocks deriving on path-containing structs without committing to a particular set of edge cases.

</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This release adds a `DefaultGenerator` impl for `PathBuf`, so `#[derive(Generate)]` works for structs with `PathBuf` fields and `gs::default::<PathBuf>()` returns a generator.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
-RELEASE_TYPE: minor
+RELEASE_TYPE: patch
 
-This release adds a `DefaultGenerator` impl for `PathBuf`, so `#[derive(Generate)]` works for structs with `PathBuf` fields and `gs::default::<PathBuf>()` returns a generator.
+This release implements `DefaultGenerator` for `PathBuf`.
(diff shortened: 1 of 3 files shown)


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+This release adds a `DefaultGenerator` impl for `PathBuf`, so `#[derive(Generate)]` works for structs with `PathBuf` fields and `gs::default::<PathBuf>()` returns a generator.
```
```suggestion
RELEASE_TYPE: patch

This release implements `DefaultGenerator` for `PathBuf`.
```

andrewgazelka:
nit @Liam-DeVoe on some OS Path can be non UTF-8 (superset of `text()`)

Liam-DeVoe:
I don't see a way to construct `PathBuf` with anything but a `str`, and `str` is required to be UTF-8 in rust, so I don't see how to construct such a `PathBuf`?

andrewgazelka:
> I don't see a way to construct `PathBuf` with anything but a `str`, and `str` is required to be UTF-8 in rust, so I don't see how to construct such a `PathBuf`?

note `Path::to_str` returns `Option<&str>`. To construct a non str based path you must go through `std::ffi::OsStr:from_bytes` and then `Path::new(os_str)`. whether OsStr is UTF-8 or not is OS dependent.

Liam-DeVoe:
Huh, interesting. https://github.com/hegeldev/hegel-rust/issues/216

PR #228, opened 2026-05-01
Add native backend

DRMacIver:
Draft state, in the process of being finalized, pull request open for CI mainly. Please ignore for now.

PR #229, opened 2026-05-01
Fix HegelValue deserializer to preserve -0.0 sign bit

DRMacIver:
<summary>Details</summary>
When deserializing `HegelValue::Number(-0.0)`, the integer-optimization branch in `deserialize_any` was triggering because `-0.0.fract() == 0.0` and `-0.0` is in the i64 range. This caused the visitor to receive `visit_i64(0)`, silently dropping the negative sign bit.

The fix adds an `is_neg_zero` check before the integer cast: if `n == 0.0 && n.is_sign_negative()`, the value bypasses the `visit_i64` path and goes through `visit_f64` instead, preserving the sign bit.

This manifested as `floats(max_value=-0.0)` generating `0.0` instead of `-0.0`, and `minimal(floats(), is_sign_negative)` shrinking to `-1.0` instead of `-0.0` because the server's `-0.0` attempts arrived on the client as `0.0` and failed the predicate.

Three new tests are added covering the negative-zero f64 path, the negative-zero f32 path, and a regression test for positive zero (confirming it still goes through the i64 path correctly).
</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
 RELEASE_TYPE: patch
 
-This patch bumps the minimum supported protocol version to take into account recent changes to `one_of`.
+This patch fixes `HegelValue` deserializing `-0.0` as `0.0` for float targets. The integer-optimisation branch was triggering for negative zero (because `-0.0.fract() == 0.0`), so the visitor received `visit_i64(0)` and the sign bit was silently lost. This caused `floats(max_value=-0.0)` to generate `0.0` instead of `-0.0`, and made `minimal(floats(), |x| x.is_sign_negative())` shrink to `-1.0` rather than `-0.0`.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,7 @@ jobs:
           MIN_CORE: ${{ steps.min-version.outputs.min_core }}
         run: |
           uv tool install "hegel-core==$MIN_CORE"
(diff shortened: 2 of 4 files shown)


Liam-DeVoe:
We should add a check for this to the conformance test as well. Might be a bit tricky, not sure the best way to do that. We probably want `FloatsConformance` to require the client to serialize its generated floats as their 8 byte binary representation.

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes `HegelValue` deserializing `-0.0` as `0.0` for float targets. The integer-optimisation branch was triggering for negative zero (because `-0.0.fract() == 0.0`), so the visitor received `visit_i64(0)` and the sign bit was silently lost. This caused `floats(max_value=-0.0)` to generate `0.0` instead of `-0.0`, and made `minimal(floats(), |x| x.is_sign_negative())` shrink to `-1.0` rather than `-0.0`.
```
DRMacIver, line 3: This release message is almost entirely about internals, and should be entirely not about internals. The specific thing that changes is how generation and shrinking of negative float bounds works in hegel tests, and it should say that.

DRMacIver (commented):
As well as the comment about the release message, this adds zero tests about the behaviour of public facing APIs, and it should add at least one - e.g. that when generating floats with max_value(-0.0) the sign should always be negative.

echoumcp1:
We may end up writing a custom deserializer here https://github.com/hegeldev/hegel-rust/pull/160

I can fold this change into it.

DRMacIver:
I'm going to roll this into #236

PR #235, opened 2026-05-01
Fix unique vecs producing duplicates when element generator is sampled_from

DRMacIver:
<summary>Details</summary>

`VecGenerator::as_basic()` was sending a list schema with `"unique": true` to the server, which enforced uniqueness on raw CBOR values (indices for `sampled_from`), not on actual parsed values. For `sampled_from(vec![0; 100])`, distinct indices 5 and 42 both map to value `0`, so the server's uniqueness check passes while the resulting vec contains duplicates.

The fix makes `as_basic()` return `None` when `unique_by` is set, forcing the non-basic `Collection` path where uniqueness is checked on actual `T` values after parsing. The `"unique": false` change is safe because the fast path only runs when `unique_by` is `None`.

A regression test `test_vec_unique_sampled_from_no_duplicates` in `tests/test_collections.rs` verifies that `vecs(sampled_from(vec![0; 100])).unique(true)` never produces a vec with more than 1 element (since all 100 elements are equal, a truly unique vec can hold at most one of them).
</details>

### DIFF
--- a/src/generators/collections.rs
+++ b/src/generators/collections.rs
@@ -74,11 +74,14 @@ where
         if let Some(max) = self.max_size {
             assert!(self.min_size <= max, "Cannot have max_size < min_size");
         }
+        if self.unique_by.is_some() {
+            return None;
+        }
         let basic = self.elements.as_basic()?;
 
         let mut schema = cbor_map! {
             "type" => "list",
-            "unique" => self.unique_by.is_some(),
+            "unique" => false,
             "elements" => basic.schema().clone(),
             "min_size" => self.min_size as u64
         };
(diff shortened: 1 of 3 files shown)


src/generators/collections.rs
```diff
@@ -74,11 +74,14 @@ where
         if let Some(max) = self.max_size {
             assert!(self.min_size <= max, "Cannot have max_size < min_size");
         }
+        if self.unique_by.is_some() {
```
DRMacIver, line 77: I think this is probably too aggressive and we can still be basic as long as min_size <= 1

PR #236, opened 2026-05-01
Reject sign-aware empty float range min=+0.0 max=-0.0 with clear error

DRMacIver:
<summary>Details</summary>

## Summary

Under IEEE 754, `+0.0 == -0.0`, so a naive `min <= max` check passes when `min=+0.0, max=-0.0`. However, Hypothesis and the native backend use sign-aware ordering (matching `bad_zero_bounds` in Hypothesis's `strategies/_internal/numbers.py` and `sign_aware_lte` in `native/core/choices.rs`), under which `-0.0 < +0.0`. This means the range `[+0.0, -0.0]` contains no valid floats.

Previously the server-side `FloatGenerator` did not detect this case, causing a silent mismatch with the native backend.

## Changes

- Adds a `to_f64()` method to the `Float` trait so the validation code can do cross-width comparisons without generic bound complications.
- Adds a `bad_zero_bounds` check in `FloatGenerator::build_schema()` that panics with a clear `InvalidArgument` message when `min=+0.0` and `max=-0.0`.
- Adds two `#[should_panic]` tests in `tests/test_validation.rs` covering both `f64` and `f32`.
- Updates the `fuzz_floats_bounds` fuzz test to skip this now-rejected case so it doesn't spuriously fail.

</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes `gs::floats()` accepting `min_value=0.0` and `max_value=-0.0` without error. Although `+0.0 == -0.0` under IEEE 754, Hypothesis and the native backend use sign-aware ordering where `-0.0 < +0.0`, so this range contains no valid floats. The generator now panics with a clear `InvalidArgument` message in this case, matching the behaviour of the native backend.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,6 @@
 RELEASE_TYPE: patch
 
-This patch fixes `gs::floats()` accepting `min_value=0.0` and `max_value=-0.0` without error. Although `+0.0 == -0.0` under IEEE 754, Hypothesis and the native backend use sign-aware ordering where `-0.0 < +0.0`, so this range contains no valid floats. The generator now panics with a clear `InvalidArgument` message in this case, matching the behaviour of the native backend.
+This patch fixes two related issues with negative zero in the float generator:
+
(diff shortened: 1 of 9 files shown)


src/generators/numeric.rs
```diff
@@ -167,6 +175,20 @@ impl<T: Float + serde::Serialize> FloatGenerator<T> {
...
+            let min_f = min.to_f64();
+            let max_f = max.to_f64();
+            if min_f == 0.0 && max_f == 0.0 && min_f.is_sign_positive() && max_f.is_sign_negative()
+            {
+                panic!(
+                    "InvalidArgument: There are no {width}-bit floating-point \
+                     values between min_value=0.0 and max_value=-0.0"
+                );
+            }
         }
```
Liam-DeVoe, line 203: I'd like to write a `sign_aware_lte` helper instead of this, as we have in hypothesis. We're going to want this logic in multiple places

RELEASE.md
```diff
@@ -0,0 +1,6 @@
+RELEASE_TYPE: patch
+
+This patch fixes two related issues with negative zero in the float generator:
+
+- `gs::floats()` previously accepted `min_value=0.0` and `max_value=-0.0` without error. Although `+0.0 == -0.0` under IEEE 754, Hypothesis and the native backend use sign-aware ordering where `-0.0 < +0.0`, so this range contains no valid floats. The generator now panics with a clear `InvalidArgument` message in this case, matching the behaviour of the native backend.
+- `HegelValue` was deserializing `-0.0` as `0.0` for float targets. The integer-optimisation branch was triggering for negative zero (because `-0.0.fract() == 0.0`), so the visitor received `visit_i64(0)` and the sign bit was silently lost. This caused `floats(max_value=-0.0)` to generate `0.0` instead of `-0.0`, and made `minimal(floats(), |x| x.is_sign_negative())` shrink to `-1.0` rather than `-0.0`.
```
Liam-DeVoe, line 6: this is exposing a bunch of internal details to readers :/ `minimal`, `visit_i64`

DRMacIver: Ugh, sorry, I reviewed the code but forgot to rereview the realease file.

PR #238, opened 2026-05-01
Add gs::nan_floats() generator for NaN-specific float generation

DRMacIver:
<summary>Details</summary>

Adds a `NanFloatGenerator` struct and `gs::nan_floats()` factory function to `src/generators/numeric.rs`, exported via `src/generators/mod.rs`.

This is a port of Hypothesis's `NanStrategy`: it draws a sign bit (boolean) and a 52-bit mantissa (integer in `[0, 2^52 - 1]`), then ORs them together with `f64::NAN.to_bits()` (which has the quiet bit set). The result is always a valid NaN, and all four variants (positive/negative × quiet/signaling) are routinely reachable within a small number of draws.

**Why this is useful**: `gs::floats()` with default settings can generate NaN, but only rarely — the probability of hitting a specific bit-pattern variant is low. `nan_floats()` is intended for tests that specifically need NaN values and want to distinguish variants (e.g. finding a negative NaN or a NaN with a non-standard mantissa).

**Tests added** in `tests/test_floats.rs`:
- `always_nan`: property test asserting all values from `nan_floats()` are NaN
- `can_find_positive_quiet_nan`, `can_find_negative_quiet_nan`, `can_find_positive_signaling_nan`, `can_find_negative_signaling_nan`: four find-any tests covering all (sign × mantissa-pattern) variants
</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
-This patch bumps the minimum supported protocol version to take into account recent changes to `one_of`.
+This release adds `gs::nan_floats()`, a generator for NaN `f64` values with varied sign bits and mantissa bit-patterns. This is a port of Hypothesis's `NanStrategy` and is useful for tests that need to distinguish between different NaN variants.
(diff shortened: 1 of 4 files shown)


DRMacIver:
Not useful on its own. I'll pull this in as part of the native work.

PR #255, opened 2026-05-12
Generate snake case enum helper methods

andrewgazelka:
## Why this is needed

AI: codex gpt 5.5 high

The DefaultGenerator derive currently emits enum helper identifiers based directly on variant names, such as Circle, default_Circle, and internal bindings like basic_Circle. Downstream crates that deny non_snake_case need crate-level lint suppressions just to derive generators for ordinary Rust enums.

## What changed

- convert generated enum generator fields, helper methods, and internal basic-generator bindings to snake_case using heck
- keep hidden PascalCase compatibility aliases so existing callers keep compiling without deprecation warnings
- handle keyword-shaped variant helpers like Type, Gen, and Crate
- update derive tests to use circle, default_circle, etc.
- add a downstream compile test with deny(deprecated, non_snake_case)

## Validation

- cargo test --test test_derive
- cargo test test_enum_derive_compiles_with_non_snake_case_denied --test test_derive_compile
- cargo clippy --all-targets --all-features -- -D warnings

I also ran cargo test --all-targets --all-features; it reached the examples and failed on the intentionally failing examples/die_hard.rs property (assertion failed: self.big != 4), unrelated to this macro change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,6 +231,7 @@ dependencies = [
 name = "hegeltest-macros"
 version = "0.12.3"
 dependencies = [
+ "heck",
  "proc-macro2",
  "quote",
  "syn",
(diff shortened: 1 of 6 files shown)


Liam-DeVoe:
funnily enough I'm working on https://github.com/hegeldev/hegel-rust/pull/154 as we speak :) 

thanks for this PR, it has some things I hadn't thought of, like keyword collision detection. I'll fold in the useful stuff here into https://github.com/hegeldev/hegel-rust/pull/154

andrewgazelka:
impl in #154

PR #256, opened 2026-05-12
Share agent guidance across Claude and Codex

andrewgazelka:
## Why this is needed

The repository currently keeps coding-agent guidance under `.claude/CLAUDE.md`, which works for Claude Code but leaves Codex 5.5 without the standard root `AGENTS.md` entry point. This PR makes the instructions shared across both tools without duplicating content.

## What changed

- add root `AGENTS.md` as the canonical agent guidance file
- move `.claude/skills` to `.agents/skills`
- make `.claude` a symlink to `.agents`
- make `.agents/CLAUDE.md` a symlink to `../AGENTS.md`, so `.claude/CLAUDE.md` still resolves for Claude compatibility
- add a release note for the repository metadata change

## Validation

- verified `.claude -> .agents`
- verified `.agents/CLAUDE.md -> ../AGENTS.md`
- verified `.claude/CLAUDE.md` resolves to the same content as `AGENTS.md`
- `python .github/scripts/release.py check main`
- `git diff --cached --check`


DRMacIver:
I would like for this to work, but my impression is that it doesn't, and would like to do some careful testing first. My understanding is that claude code often completely fails to read CLAUDE.md when it's a symlink, and tends to work much worse with this sort of setup.

This might be out of date knowledge, but we're sufficiently heavy claude code users that I'm hesitant to do things that make that worse.

andrewgazelka:
> I would like for this to work, but my impression is that it doesn't, and would like to do some careful testing first. My understanding is that claude code often completely fails to read CLAUDE.md when it's a symlink, and tends to work much worse with this sort of setup.
> 
> This might be out of date knowledge, but we're sufficiently heavy claude code users that I'm hesitant to do things that make that worse.

I use both codex and claude and do not see this occur anymore. we can also reverse the symlink direction... at face value really annoying to make PRs when OAI releases new models right now though. I always like using what is SOTA

DRMacIver:
Fair enough! Well, I'm happy to give it a try, and if it causes us problems we can try something else. It'll only affect our development workflows, so it'll be easy to see if it's a problem.

Liam-DeVoe:
I'm also happy to give this a try

PR #259, opened 2026-05-13
docs: hegel for rust - line edit + request to define terms

ink7047:
I recommend defining all terms in the "modules" section for consistency and completeness (see in-line comments)


DRMacIver:
Thanks! Agreed the copy-edit is an improvement.

For the modules, I think we should actually be hiding `backend` and `explicit_test_case` from docs. The former is internal only, and the latter is only for use by the re-exported attribute macro. `extras` definitely needs more explanatory docs though (it's a catch-all namespace for our integration with various external libraries).

Liam-DeVoe (approved):
Thanks!

PR #262, opened 2026-05-13
Minimal implementation of hegel native feature set

DRMacIver:
This tries to be a minimal native feature, extracted from #228 and heavily cleaned up. The large PR was getting too unwieldy and contained too many lurking problems, so I'm now going to try to basically treat it as an artefact of study and merge one PR at a time (though still on a fairly aggressive schedule) of code that is at least reviewable if not as thoroughly reviewed as I'd like.


src/native/shrinker/mutation.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Results with more than this many nodes are skipped.
+const MAX_MUTATE_NODES: usize = 32;
+
+impl<'a> Shrinker<'a> {
+    /// Try random mutations of a few positions to escape local optima.
+    ///
+    /// Port of pbtkit's `shrinking/mutation.py::mutate_and_shrink`.
+    pub(super) fn mutate_and_shrink(&mut self) {
+        if self.current_nodes.len() > MAX_MUTATE_NODES {
+            return;
```
Liam-DeVoe, line 29: it's hard for me to mentally map this to Hypothesis concepts when it uses pbtkit terminology and, seemingly, differing semantics. It seems this pass is similar to `reduce_each_alternative`, but not restricted to small integer choices.

If we try to change the shrinker semantics at the same time we're porting the shrinker, we're going to have a bad time attributing any behavior differences. I'd prefer if we only ported Hypothesis, and didn't simultaneously add anything new.

src/native/database.rs
```diff
@@ -0,0 +1,273 @@
...
+        self.db_root.join(fnv_hex(key))
+    }
+
+    fn value_path(&self, key: &[u8], value: &[u8]) -> PathBuf {
+        self.key_path(key).join(fnv_hex(value))
+    }
+}
+
+impl ExampleDatabase for NativeDatabase {
+    fn fetch(&self, key: &[u8]) -> Vec<Vec<u8>> {
```
Liam-DeVoe, line 110: `Example` should be a banned term for claude to use without consulting a human :)

src/native/shrinker/mutation.rs
DRMacIver: This is a reasonable preference for you to have, but I'm going to ignore it, sorry. Happy to talk about it more offline but the overriding reason is at this point what I really want to do is get this working based on the large existing artefact of mostly good and working ported code that we already have, and making further large changes to that is going to slow down a process that is already weeks behind where I'd like it to be at this stage.

src/native/database.rs
DRMacIver: It should. In its defence this is straight up ported from Hypothesis. We should come up with a better name but it's internal so I will probably not block merging on that.

src/native/shrinker/mutation.rs
Liam-DeVoe: OK. Understand that this will make my ability to review these pull requests worse

DRMacIver: Acknowledged. I know we're taking on significant review debt with this work, and I think that's OK.

src/native/det_tree.rs
```diff
@@ -0,0 +1,103 @@
...
+//! [`super::test_runner`]: each had its own `ChoiceValueKey` enum, its own
+//! trie struct (`TreeNode` / `DetTreeNode`), and its own copy of the
+//! divergent-kind panic wording. Wording drift between the two caused
+//! T0.1; structural drift would have caused more. This module is the
+//! single source of truth for both runners.
+//!
+//! Note: `super::conjecture_runner` has a *separate*, richer trie
+//! ([`super::conjecture_runner::DataTreeNode`]) that additionally tracks
+//! conclusion status, kill depths, and exhaustion for novel-prefix
+//! generation. It is *not* unified here — its bookkeeping is specific to
```
DRMacIver: This seems weird and entirely wrong. What on earth is this for? Why is there an entirely separate tree? That's not how it works in anything that was meant to be ported, and this entire module seems strictly unnecessary.

src/native/mod.rs
```diff
@@ -0,0 +1,19 @@
+// Native pbtkit-style test engine for Hegel.
```
DRMacIver: We really shouldn't be littering the code with references to pbtkit, and also this should be more Hypothesis than pbtkit.

src/native/data_source.rs
```diff
@@ -0,0 +1,219 @@
...
+
+impl DataSource for NativeDataSource {
+    fn generate(&self, schema: &Value) -> Result<Value, DataSourceError> {
+        use crate::cbor_utils::cbor_map;
+        self.dispatch("generate", &cbor_map! {"schema" => schema.clone()})
+    }
+
+    fn start_span(&self, label: u64) -> Result<(), DataSourceError> {
+        use crate::cbor_utils::cbor_map;
+        self.dispatch("start_span", &cbor_map! {"label" => label})?;
```
DRMacIver: Seems very unnecessary for these to go through a dispatch loop. Generating from a schema needs to because that's an actual thing that needs interpreting, but why would these?

src/native/database.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn key_path(&self, key: &[u8]) -> PathBuf {
+        self.db_root.join(key_hash(key))
+    }
+
+    fn value_path(&self, key: &[u8], value: &[u8]) -> PathBuf {
+        self.key_path(key).join(fnv_hex(value))
+    }
+}
+
+impl ExampleDatabase for NativeDatabase {
```
DRMacIver: We should probably call this TestCaseDatabase in Hegel.

src/native/database.rs
```diff
@@ -0,0 +1,268 @@
...
+            self.delete(src, value);
+            self.save(dst, value);
+            return;
+        }
+        let _ = std::fs::remove_dir(self.key_path(src));
+    }
+}
+
+/// FNV-1a 64-bit hash of a byte slice, formatted as a 16-character hex string.
+pub(super) fn fnv_hex(s: &[u8]) -> String {
```
DRMacIver, line 192: It would be useful to have a comment about why we need this.

src/native/tree.rs
```diff
@@ -0,0 +1,30 @@
+//! [`NativeRunner`] + [`RunResult`] surface used by the shrinker.
```
DRMacIver: Seems weird to have these in a tiny shim module named `tree`

tests/embedded/server/data_source_tests.rs
```diff
@@ -1,16 +1,186 @@
...
+            }
+        }
+    });
+
+    ServerDataSource::new(conn, stream, Verbosity::Quiet).0
+}
+
+// ── N8: ServerDataSource::target_observation client-side validation ───────
+//
+// The audit (item N8) flagged that `ServerDataSource::target_observation`
```
DRMacIver: Do not reference transient things like audit items in comments.

tests/embedded/run_lifecycle_tests.rs
```diff
@@ -0,0 +1,168 @@
+//! Embedded tests for `src/run_lifecycle.rs`.
+
+use super::*;
+
+// ── N18.run_lifecycle: panic_message fallback for non-string payloads ────
```
DRMacIver: Labels like `N18.run_lifecycle` make no sense and should be omitted.

src/server/process.rs
```diff
@@ -127,13 +127,18 @@ fn version_check_message(
...
                 Some(format!(
                     "Version mismatch: expected '{}', got '{}'.",
                     expected_version_string, stdout
                 ))
             } else {
-                None
+                Some(format!(
+                    "'{}' --version returned unexpected output. Is this a hegel binary?",
+                    binary_path
+                ))
```
DRMacIver: This seems fine, but I'm not sure why this change was needed.

src/server/session.rs
```diff
@@ -372,12 +371,12 @@ impl TestRunner for ServerTestRunner {
...
                         .expect("Failed to ack test_case");
 
-                    let backend = Box::new(ServerDataSource::new(
-                        Arc::clone(connection),
-                        test_case_stream,
-                        verbosity,
-                    ));
-                    run_case(backend, false);
+                    if verbosity == Verbosity::Verbose {
+                        eprintln!("Trying example: ");
```
DRMacIver: The word "example" is banned from Hegel without very careful thought. I would have this say "Running test case"

PR #263, opened 2026-05-13
tests: fix running on NixOS

winterqt:


Liam-DeVoe:
Bleh, apparently rust doesn't have a way to conditionally skip a test based on a variable? You can only conditionally skip based on a feature?

I'm ok with merging this, but let's extract the nixos detection to a shared `is_nixos` variable, probably in `tests/common`.

PR #265, opened 2026-05-13
hegel-macros: support async `#[hegel::test]`s

winterqt:


hegel-macros/src/hegel_test.rs
```diff
@@ -47,17 +78,30 @@ pub fn expand_test(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        return syn::Error::new_spanned(
+            asy,
+            "#[hegel::test] used on an async function; move #[hegel::test] below your async test macro (like #[tokio::test]).",
+        )
+        .to_compile_error();
+    }
+
     for attr in &func.attrs {
+        // We special case the common case of "is the user trying to write this?" Other macros, like `#[tokio::test]`,
+        // don't use the bare `test` identifier from prelude, instead being caught by `is_test_attribute`.
```
winterqt: Happy to discuss this further as it's not what we agreed upon, but I'm pretty sure it's the path of least resistance.

DRMacIver: I'm basically fine with it as a solution, but I find it a little weird that we are making this a compile error and making all other test attributes silently work correctly.

I'm basically fine with either of those choices, but the inconsistency between the two bugs me a little.

Liam-DeVoe: Yes the inconsistency here is bad.

I've fixed this by removing this special case, so that `#[hegel::test]` always works with `#[test]` (when we detect it).

(I've also added a bunch of tests, which might interest you @winterqt if you have any personal interest in what I consider to be high-quality tests for a pr)

Liam-DeVoe (approved):
Thanks :)

hegel-macros/src/hegel_test.rs
winterqt: Thanks, that seems like a slightly nicer solution, I just didn’t want to remove compile errors.

re: tests, I omitted them on purpose: we didn’t have tests for some of the macro behaviors (like the aforementioned error message), and I also didn’t know if you wanted a dependency to be added (tokio) just for some macro expansion test.

Do we have documentation on when we want new tests to be added for new/changed functionality?

winterqt: (The `CONTRIBUTING.md` is very small at the moment. :))

winterqt:
That’s an interesting test failure. Let me poke at my test case I used for making this and see why this is only failing in the minimum versions case.

hegel-macros/src/hegel_test.rs
Liam-DeVoe: yeah that's very valid on your end. We definitely should have had a test for the error message before, and didn't. As for dependencies, I'm pretty happy to pull in dependencies when we need to test our explicit integration with those dependencies. I think eventually we'll split our `dev-dependencies` in two to section off all the one-off dependencies used for one or two tests into their own CI job, to speed up build times in the common case.

Liam-DeVoe:
Claude came back with a good explanation: https://github.com/hegeldev/hegel-rust/pull/265/commits/d02ef315a267e27d2ddce648e0b30afd64f777ee. Only one of the errors was load-bearing, the others were incidental warnings.

Liam-DeVoe:
(ok, clearly not that good if it failed again. we'll see where claude tracks it down to this time.)

Liam-DeVoe:
Claude's analysis:

<details>

⏺ Found the real root cause. tokio-macros 1.7.0 does accept args, but it prepends #[::core::prelude::v1::test] to the attribute list. That means libtest's
  #[test] processing sees the args-having function before #[hegel::test] gets to strip the args. tokio PR #6497 (May 2024) fixed this by appending the test
  attribute to the end of the attribute list instead — that way #[hegel::test] runs first and cleans up before libtest looks at the signature.

  That fix is in tokio-macros 2.4.0, which tokio 1.39.0 is the first release to require (via ~2.4.0). So the real minimum is tokio = "1.39". Let me update
  and verify.

⏺ Update(/Users/tybug/Desktop/Liam/coding/hegel/rust/Cargo.toml)
  ⎿  Added 1 line, removed 1 line
      62  # Note: tokio 1.15 is when tokio started forwarding test parameters on #[tokio::test] to
      63  # the macro-expanded function, which is a requirement for #[hegel::test] to continue
      64  # accepting TestCase.
      65 -tokio = { version = "1.15", features = ["macros", "rt"] }                                                                                   
      65 +tokio = { version = "1.39", features = ["macros", "rt"] }                                                                                   
      66  # `hegeltest-macros` is a regular dep of hegel itself, but it's also listed
      67  # here with the `internal-test-only` feature so the integration-test suite
      68  # can pull in macros that are intentionally hidden from external users

⏺ Bash(cargo +nightly generate-lockfile -Z minimal-versions 2>&1 | tail -2 && grep -A2 '^name = "tokio"$' Cargo.lock | head -10)
  ⎿        Adding wit-bindgen-core v0.51.0 (requires Rust 1.87.0)
           Adding wit-bindgen-rust v0.51.0 (requires Rust 1.87.0)
     name = "tokio"
     version = "1.39.1"
     source = "registry+https://github.com/rust-lang/crates.io-index"

⏺ Bash(grep -A2 '^name = "tokio-macros"$' Cargo.lock | head -10 && echo "---" && cargo test --locked --features
      rand,antithesis,chrono,jiff,serde_json,serde_json_raw_value --test test_async --no-run 2>&1 | tail -20)
  ⎿  name = "tokio-macros"
     version = "2.4.0"
     source = "registry+https://github.com/rust-lang/crates.io-index"
     ---
        Compiling pin-project-lite v0.2.11
        Compiling tokio v1.39.1
        Compiling hegeltest v0.13.0 (/Users/tybug/Desktop/Liam/coding/hegel/rust)
         Finished `test` profile [unoptimized + debuginfo] target(s) in 1.31s
       Executable tests/test_async.rs (target/debug/deps/test_async-91f1cd3c6a768c8a)

⏺ Bash(cargo test --locked --features rand,antithesis,chrono,jiff,serde_json,serde_json_raw_value --test test_async 2>&1 | tail -15)
  ⎿      Finished `test` profile [unoptimized + debuginfo] target(s) in 0.04s
          Running tests/test_async.rs (target/debug/deps/test_async-91f1cd3c6a768c8a)

     running 2 tests
     test test_async_with_tokio ... ok
     test test_hegel_test_on_async_fn_without_wrapper_fails ... ok

     test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.97s

⏺ Verified locally. With the lockfile minimal-versions-resolved:

  - tokio v1.39.1
  - tokio-macros v2.4.0

  Both tests pass.

  Updated diagnosis

  tokio 1.15/tokio-macros 1.7 was NOT enough. Two separate tokio fixes were needed:

  1. tokio-macros 1.2.0 (April 2021, tokio 1.5+ via path dep, transitively required by tokio 1.15+) — forward #[tokio::test]'s function arguments instead of
  erroring on them. (PR #3691)
  2. tokio-macros 2.4.0 (May 2024, transitively required by tokio 1.39+) — append the generated #[::core::prelude::v1::test] to the end of the attribute list
   rather than prepending it. (PR #6497)

  Without fix #2, even though tokio happily forwards tc: TestCase through, libtest's #[test] handling sees the args-having function before #[hegel::test]
  gets a chance to strip them, and errors with "functions used as tests can not have any arguments". With #2, #[hegel::test] runs first (it's earlier in the
  attribute list), clears args, and then #[::core::prelude::v1::test] sees the clean signature.

  Fix

  Bumped tokio dev-dep from "1" → "1.39" in Cargo.toml:62.

  Release dates:
  - tokio 1.0.0: 2020-12-23
  - tokio 1.39.0: 2024-07-23

  So this was a ~3.5-year-old tokio requirement. MSRV-wise tokio 1.39 lists rust-version = "1.70", still well below this crate's 1.86 floor.

</details>

PR #270, opened 2026-05-14
Add native support for bytes type

DRMacIver:
<details>
<summary>LLM-generated implementation notes</summary>

This PR was extracted from `DRMacIver/native` by Claude Code following the `landing-native-chunk` skill.

## What this adds

- `BytesChoice { min_size, max_size }` in `src/native/core/choices.rs`, with `simplest`, `unit`, `validate`, `sort_key`, and index round-trip (BigUint shortlex). Refactored `NodeSortKey` to an enum (`Scalar` vs `Sequence`) so byte-string nodes can sort by `(len, lex)`.
- `draw_bytes` on `NativeTestCase`, with `biased_bytes_sample` and a many-state-driven length draw mirroring the float chunk's structure.
- `interpret_binary` schema interpreter (`src/native/schema/bytes.rs`); wired into the schema dispatch — `binary` schemas no longer `todo!()` under native.
- Database tag `3` for `Bytes` in the choice serialization format: 4-byte LE length followed by raw bytes, with bounds-checked deserialization.
- `Bytes(Vec<u8>)` arm added to `ChoiceValueKey` in `data_tree.rs`.
- Two shrinker passes in `src/native/shrinker/bytes.rs`: `shrink_bytes` (simplest, binary-search shorten, linear-scan fallback, delete, reduce-toward-0, insertion-sort) and `redistribute_bytes_pairs` (move bytes between adjacent `BytesChoice` pairs). Wired into the shrink loop after the float passes.
- `is_float_or_integer` in the float shrinker now has a `Bytes` arm returning `false`.

## What this lets through

`#[not_supported_on_native]` is dropped from every test that was gated only because of a `binary` schema draw — including `tests/rand/randoms.rs::test_randoms_fill`, and several places in `tests/test_quality.rs`, `tests/test_collections.rs`, `tests/test_hegel_test.rs`, `tests/test_shrink_quality/bytes.rs`, `tests/test_shrink_quality/composite.rs`.

## What this doesn't add

No new public API (no `gs::bytes()` generator). This is purely engine-level: server-mode tests that draw bytes from Hypothesis can now replay under `--features native`. A `bytes()` generator can land in a follow-up.

## Test plan

- [x] `just check` passes (clippy + fmt + native test suite + server test suite + doctests)
- [x] `just check-coverage` clean (0 uncovered code lines)
- [x] 3 stability runs of `cargo test --features native`: 0 failures each
- [x] `cargo test` (server suite): 0 failures
- [ ] Watch CI on the open PR

</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,8 +1,3 @@
 RELEASE_TYPE: patch
 
 The native backend (`--features native`) now supports byte-string generators.
-The `binary` schema (`min_size` / `max_size` bounds) is interpreted natively,
-and the shrinker has bytes-specific passes (`shrink_bytes` for shortening
-and lowering, `redistribute_bytes_pairs` for moving length between adjacent
-bytes nodes). Tests that draw bytes via Hypothesis schemas can now replay
-under `--features native`, including `tests/rand/randoms.rs::test_randoms_fill`.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,8 @@
+RELEASE_TYPE: patch
+
+The native backend (`--features native`) now supports byte-string generators.
+The `binary` schema (`min_size` / `max_size` bounds) is interpreted natively,
+and the shrinker has bytes-specific passes (`shrink_bytes` for shortening
+and lowering, `redistribute_bytes_pairs` for moving length between adjacent
+bytes nodes). Tests that draw bytes via Hypothesis schemas can now replay
+under `--features native`, including `tests/rand/randoms.rs::test_randoms_fill`.
--- a/src/native/core/choices.rs
+++ b/src/native/core/choices.rs
@@ -555,7 +555,7 @@ pub enum ChoiceValue {
 }
 
 /// Bit-exact equality for floats keeps `-0.0` distinct from `0.0` and
-/// preserves NaN payloads; integers, booleans and bytes use natural equality.
+/// preserves NaN payloads; other choice types use natural equality.
 impl PartialEq for ChoiceValue {
     fn eq(&self, other: &Self) -> bool {
         match (self, other) {
--- a/src/native/core/choices.rs
+++ b/src/native/core/choices.rs
@@ -177,6 +177,97 @@ impl BooleanChoice {
     }
 }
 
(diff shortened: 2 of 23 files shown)


src/native/core/choices.rs
```diff
@@ -459,16 +551,18 @@ pub enum ChoiceValue {
     Integer(i128),
     Boolean(bool),
     Float(f64),
+    Bytes(Vec<u8>),
 }
 
 /// Bit-exact equality for floats keeps `-0.0` distinct from `0.0` and
-/// preserves NaN payloads; integers and booleans use natural equality.
+/// preserves NaN payloads; integers, booleans and bytes use natural equality.
```
DRMacIver: Nit: I'd just say "other choice types" here rather than enumerating them all. Floats are the only ones that are weird.

src/native/core/state.rs
```diff
@@ -96,6 +96,27 @@ static GLOBAL_CONSTANTS_INTEGERS: LazyLock<Vec<i128>> = LazyLock::new(|| {
...
+/// distribution: length clusters around a small `average_size` derived
+/// from `min(max(min_size * 2, min_size + 5), 0.5 * (min_size + max_size))`.
+///
+/// Hypothesis: `conjecture/providers.py::HypothesisProvider.draw_bytes`
+/// (and `draw_string`).
+fn many_draw_length(rng: &mut SmallRng, min_size: usize, max_size: usize) -> usize {
+    if min_size == max_size {
+        return min_size;
+    }
+    let many = ManyState::new(min_size, Some(max_size));
```
DRMacIver: We should extract the p_continue calculation rather than creating a ManyState just to get it.

Also we should calculate this size using a standard geometric variate variate algorithm here, so this is `min(max_value, min_value + floor(log(random())/log(p_continue))`

RELEASE.md
```diff
@@ -0,0 +1,8 @@
+RELEASE_TYPE: patch
+
+The native backend (`--features native`) now supports byte-string generators.
```
DRMacIver, line 3: This sentence is really all the release needed to include. The rest is redundant.

PR #280, opened 2026-05-19
Port Hypothesis #4728: new integer distribution

DRMacIver:
This is an automated port of https://github.com/HypothesisWorks/hypothesis/pull/4728

To the best of my ability to validate it looks sane, but I didn't understand the original work well enough that I feel fully confident in that assesment. 

<details>
<summary>Implementation notes (LLM-generated)</summary>

Ports [HypothesisWorks/hypothesis#4728](https://github.com/HypothesisWorks/hypothesis/pull/4728) (the "new integers distribution" change Liam landed upstream) to the native backend.

**What it replaces.** `biased_integer_sample` in `src/native/core/state.rs` used to fall back to `rng.random_range(min..=max)` whenever the boundary-boost branch didn't fire. For wide ranges (e.g. `[i64::MIN, i64::MAX]`) that put almost all probability mass in the top few bits, identical to the pre-#4728 Hypothesis behaviour.

**What it does instead.** A new `INTEGERS_DISTRIBUTION` static — a piecewise splice of `UniformDistribution(±256)` (inner core) and `LogStudentTDistribution(scale_bits=13, df=2)` (heavy outer tails) — is restricted to the requested `[min_value, max_value]` via its CDF, then inverse-CDF sampled. Magnitudes now spread smoothly across decades. If the CDF window across the bounds is narrower than 1e-13, the code falls back to uniform (matching Hypothesis's safety branch).

**Vendored math.** `src/native/statistics.rs` is a direct port of `hypothesis-python/src/hypothesis/internal/statistics.py`:
- `stdtr` — Student's t CDF, A&S 26.7.7-8 finite sum.
- `stdtrit` — Student's t quantile; closed-form Shaw quantile for df ∈ {1, 2}, bracketed Newton iteration on `stdtr` for df ≥ 3.
- `UniformDistribution`, `LogStudentTDistribution`, `PiecewiseDistribution` — the same `Distribution` trait shape as Liam's Python `_Distribution`.
- A small Lanczos `ln_gamma` (no libm dependency) for the Newton iteration; only ever called with `x ≥ 1.0` in practice.

**Test oracle.** `tests/embedded/native/statistics_oracle_data.rs` is a static table of `(df, x, expected_scipy)` rows generated by `scripts/generate_statistics_oracle.py` (run once against scipy 1.17.1). The Rust tests assert agreement at the same tolerances Hypothesis asserts against scipy: `rel=1e-12, abs=1e-15` for `stdtr`; `rel=1e-15` for `stdtrit` with df ∈ {1, 2}; `rel=1e-7, abs=1e-9` for df ≥ 3.

**New tests.**
- 25 statistics tests (`tests/embedded/native/statistics_tests.rs`): the scipy-oracle comparisons plus symmetry/round-trip/edge-case checks for all three distributions.
- 8 sampler tests appended to `tests/embedded/native/state_tests.rs`: stay-in-range across small/wide/full-i128 bounds, collapse-when-min-equals-max, log-skewed median (the discriminating test that would fail under the old uniform fallback), unbounded-magnitude diversity, near-zero concentration, and the 1e-13 uniform-fallback path.

All 703 native lib tests and the full `cargo test --features native` suite pass. Coverage was not run locally due to disk constraints; CI will exercise it.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>
</details>


src/native/core/state.rs
Liam-DeVoe, line 212: not related to this PR, but - has claude gone wild here? this `interesting` list is nowhere in hypothesis and is going to end up upweighting these small values even more than they should be

Liam-DeVoe (approved):
nothing looks obviously wrong to me

src/native/core/state.rs
DRMacIver: Huh. I'm not sure where this came from. I suspect that happened in the original native port when it was struggling to get good results, maybe with duplicated values, and half-assed a fix.

PR #296, opened 2026-05-31
Switch the native engine's internal maps from SipHash to FxHash

DRMacIver:
<details>
<summary>What changed and why</summary>

The native engine's internal `HashMap`/`HashSet`s — the data-tree child map (`data_tree.rs`) and the span-mutation bookkeeping maps (`test_runner.rs::try_span_mutation`) — are keyed by our own choice values and span labels, never by adversarial input. The standard library's default hasher (SipHash) is DoS-resistant by design, which is pure overhead here. These maps are touched on the generation hot path (the data tree on every recorded choice; the span maps rebuilt per valid result), and SipHash showed up prominently in profiling.

This switches them to `rustc-hash`'s FxHash (added as an optional dependency gated on the `native` feature). `FxBuildHasher: Default`, so the `#[derive(Default)]` on the affected structs keeps working unchanged.

**Tests:** `data_tree` unit tests, plus `test_collections` / `test_targeting`, all pass; full native suite green.
</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -398,6 +398,7 @@ dependencies = [
  "paste",
  "rand",
  "regex",
+ "rustc-hash",
  "serde",
  "serde_json",
  "tempfile",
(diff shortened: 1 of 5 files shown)


DRMacIver:

PR #300, opened 2026-05-31
Bound regex nesting depth to avoid a native stack overflow

DRMacIver:
<details>

The native regex parser recurses (`parse`/`parse_sub`) once per nested group/alternation; only *capturing* groups were bounded (`MAXGROUPS`), so a pattern like `"(?:"*N + ")"*N` overflowed the Rust stack and aborted the process. Generation recurses over the same AST, so bounding parse depth bounds both. A new `MAX_NESTING` constant (100) makes `parse` return a clean `InvalidArgument` parse error, matching CPython's catchable `RecursionError` behaviour. Regression tests: `err_deeply_nested_groups`, `moderately_nested_groups_parse`.
</details>


DRMacIver:

PR #301, opened 2026-05-31
Reject the \z regex anchor in native to match Python re

DRMacIver:
<details>

The native regex parser mapped `\z` to an end-of-string anchor, but CPython's `re` has no `\z` (only `\Z`) and reports "bad escape \z". Removed the `\z` entry from `lookup_category` so it falls through to the bad-escape path, matching Python. Regression test: `err_z_anchor_not_supported`.
</details>


DRMacIver:

PR #302, opened 2026-05-31
Write native example-database values atomically

DRMacIver:
<details>

`DirectoryTestCaseDatabase::save` used `fs::write` directly to the value's final path, so a concurrent reader (a parallel `cargo test` binary or another process sharing `.hegel/examples`) could read a half-written value. It now writes to a `NamedTempFile` in the same directory and atomically `persist`s (renames) it into place. The value's content is fixed by its path (its FNV hash), so two processes writing the same value race harmlessly. Covered by the existing database tests under `--features native`.
</details>


DRMacIver:

PR #304, opened 2026-05-31
Report tc.target() misuse as a usage error rather than a counterexample

DRMacIver:
<details>

In the native backend, `target_observation` panicked on a non-finite score or a repeated label; the lifecycle's `catch_unwind` turned that into a `TestCaseResult::Interesting`, so a usage error was reported (and shrunk) as a found counterexample. The native data source now prefixes those panic messages with a sentinel (`TARGET_USAGE_ERROR_PREFIX`); `run_lifecycle::run_test_case` detects the prefix and `resume_unwind`s the stripped message, aborting the run with a clear error. Shared lifecycle code compiles under both backends; the regression tests are gated to `--features native`.
</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+In the native backend (`--features native`), misusing `tc.target()` (a non-finite score, or the same label twice in one test case) now aborts the run with a clear usage error instead of being caught and misreported as a failing example.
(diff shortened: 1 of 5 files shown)


DRMacIver:
Superseded by #310, which generalises this fix: instead of a `tc.target()`-specific sentinel, all invalid-argument (usage) errors — bad generator bounds, empty `sampled_from`/`one_of`, unsatisfiable filters, and `tc.target()` misuse — now share one `invalid_argument!` mechanism that aborts the run cleanly instead of being misreported (and shrunk) as a property failure. Closing in favour of #310.

PR #305, opened 2026-05-31
Report native generation non-determinism as a failure instead of aborting

DRMacIver:
<details>

`record_tree` `panic!`d on a choice-kind mismatch (a non-deterministic generator). The panic unwinds fine for the in-process library, but across the C-ABI boundary in libhegel it aborts the host process. `record_tree` now returns `Option<String>` (the diagnostic); the main generation loop folds a `Some` into a failing `TestRunResult` via `health_check_failure`, while the early trivial-probe and targeting callbacks ignore it (the main loop re-detects it). The main library still turns the failing result into a panic with the same message, so the existing `should_panic` integration test (`test_flaky_global_state`) is unchanged. Embedded test updated to `record_tree_reports_kind_mismatch`.
</details>


DRMacIver:

PR #310, opened 2026-06-01
Report invalid-argument errors uniformly as usage errors

DRMacIver:
<details>
## Summary

A *usage error* — a generator misconfigured with `max < min`, a float range that contains no values, an empty `sampled_from`/`one_of`, an unsatisfiable filter, or `tc.target()` called with a non-finite score or a repeated label — is a mistake in how the test is **written**, not a property that failed on some input.

These are detected mid-draw and panic, so the lifecycle's `catch_unwind` caught them and misreported (and pointlessly *shrank*) them as discovered counterexamples: `Property test failed: ...`. There were two separate ad-hoc conventions for this — an `InvalidArgument`-prefixed string used across the generators, and (in #304) a target-specific sentinel — neither of which actually aborted cleanly.

This unifies them into one mechanism.

## Mechanism

- `invalid_argument!` macro + `raise_invalid_argument` helper (in `test_case`). It is **context-aware**:
  - **Inside a running test body**, it panics with `INVALID_ARGUMENT_PREFIX`, which `run_lifecycle::run_test_case` strips and re-raises — aborting the run with the message, no "Property test failed", no shrinking.
  - **Outside any run** (a generator validating its arguments eagerly, e.g. `gs::sampled_from(&[])` built up front), it panics with the bare message — so the marker never leaks to the user.
- `tc.repeat()` re-raises the marker directly too.

## Routed through it

- Every draw-time argument validation in the generators (`numeric`, `strings`, `collections`, `time`, `combinators`, and the unsatisfiable-filter check).
- The `DataSourceError::InvalidArgument` arm of `panic_on_data_source_error` — which covers all `src/native/schema/**` `EngineError::InvalidArgument` sites for free.
- Both backends' `tc.target()` finite/duplicate-label checks.
- The server backend now maps `InvalidArgument` responses to `DataSourceError::InvalidArgument`, so reporting is consistent across the server and native (`--features native`) backends.

**libhegel is unaffected** — it consumes `DataSourceError::InvalidArgument` as a `Result` (→ `HEGEL_E_INVALID_ARG`) and never goes through `panic_on_data_source_error`.

## Tests

New `tests/test_usage_errors.rs` asserts the shared contract for each source — the message survives, it is **not** wrapped as a property failure, and the internal marker never leaks — on both backends.

## Supersedes #304

#304 fixed only the `tc.target()` case, with a target-specific sentinel. This generalises that fix to all invalid-argument errors, so #304 will be closed in favour of this.

</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes how invalid-argument (usage) errors are reported. Misconfiguring a generator — `max_value` below `min_value`, a float range that contains no values, an empty `sampled_from`/`one_of`, an unsatisfiable filter — or misusing `tc.target()` (a non-finite score, or the same label twice in one test case) is a mistake in how the test is written, not a property that failed. Previously such errors were caught mid-draw and misreported (and pointlessly shrunk) as a discovered counterexample ("Property test failed: ..."); now the run aborts immediately with the error message. All of these errors share a single mechanism, so the reporting is consistent across generators and across the server and native (`--features native`) backends.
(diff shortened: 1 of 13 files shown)


DRMacIver:

PR #314, opened 2026-06-01
Add #[hegel::reproduce_failure(...)]

echoumcp1:


hegel-c/include/hegel.h
```diff
@@ -390,6 +390,22 @@ void hegel_settings_database(hegel_settings_t *s, const char *database);
...
+ run — encodes a counterexample's choice sequence. When set, the engine
+ decodes it and runs exactly that one example, bypassing generation and
+ shrinking. A blob that cannot be decoded, or that no longer reproduces a
+ failure, surfaces as a failing run with an explanatory diagnostic rather
+ than a pass.
+
+ - `blob = NULL` → clear it (the default; run normally).
+ - Otherwise → replay the example encoded by `blob`.
+ */
+void hegel_settings_reproduce_failure(hegel_settings_t *s, const char *blob);
```
DRMacIver: I'm not super enamoured of this interface. We've been talking about whether we want to support multiple reproduce_failures, and I think we probably do want to allow for that possibility (e.g. your work on hegel-ocaml has multiple reproduce failures I think?) and this blocks that off.

I would maybe suggest the following interface: Have a function that takes a failure blob and returns a test case.

I also don't love that this takes a zero terminated string. I think zero terminated strings are pain (I'm not confident saying this that the rest of the API is consistent with this desire, but if it's not that's on me to fix). I'd like it to take a length prefixed string instead.

hegel-c/include/hegel.h
```diff
@@ -655,6 +671,15 @@ const char *hegel_failure_diagnostic(const hegel_failure_t *f);
...
 const char *hegel_failure_origin(const hegel_failure_t *f);
 
+/*
+ The failure's reproduce blob — a base64 string encoding the minimal
+ counterexample's choice sequence, suitable for deterministic replay via
+ `hegel_settings_reproduce_failure`. Returns NULL if `f` is NULL or the
+ engine produced no blob for this failure. The pointer is borrowed from the
+ parent `hegel_run_result_t` and stays valid until `hegel_run_free`.
+ */
+const char *hegel_failure_reproduce_blob(const hegel_failure_t *f);
```
DRMacIver: Similarly would like length-prefixing on the return. Not sure what the best API for this is. We could just define a length prefixed string type, or we could use out parameters. Whatever's easiest is fine by me.

DRMacIver (changes requested):

hegel-c/include/hegel.h
echoumcp1: I don't think it blocks multiple failure blobs for hegel-ocaml because it's implemented by calling run in a loop with a different failure blob each time. That being said the Rust version in this PR only supports a single failure blob at a time, so that should be changed.

hegel-c/include/hegel.h
echoumcp1: The rest of the library uses NUL terminated strings except the binary CBOR (schema_cbor/schema_len and
  out_value_cbor/out_value_len). I think it would be better to tackle all of the NUL terminated strings in one PR.

DRMacIver: Yeah, agreed, let's defer that to a later PR and keep this one consistent, but I would definitely like to make that change.

hegel-c/include/hegel.h
DRMacIver: I think independently of whether we actually end up supporting multiple failure blobs, I don't like blocking it off here, and I especially don't like stuffing it into the settings.

My inclination would be to have a `test_case *hegel_test_case_from_blob(hegel_settings_t *s, const char *blob)` (or similar) and let callers decide how to use it.

echoumcp1: I added the hegel_settings_t *s argument for futureproofing, but I'm pretty sure the replay with reproduce_blob doesn't use settings right now.

DRMacIver: Well it's writing its data to the settings object, which is the bit I don't like.

For creating the test case, I think you need the settings so as to propagate verbosity (which might be set to quiet)

hegel-c/include/hegel.h
```diff
@@ -457,6 +460,42 @@ const hegel_run_result_t *hegel_run_result(hegel_run_t *run);
...
+ `hegel_test_case_from_blob`. Safe to call with NULL (no-op), and safe
+ whether or not the test case was marked complete.
+
+ Must NOT be called on a test case obtained from
+ `hegel_next_test_case` — those are borrowed from the parent
+ `hegel_run_t` and are released by `hegel_run_free`. Passing one here is
+ detected (while the run is still alive) and refused, with a diagnostic
+ in `hegel_last_error_message`.
+ */
+void hegel_test_case_free(hegel_test_case_t *tc);
```
DRMacIver, line 497: Having a free that you should only sometimes call strikes me as a very bad idea. Is there a reason we can't use mark_complete like we use elsewhere here?

hegel-c/include/hegel.h
```diff
@@ -655,6 +694,15 @@ const char *hegel_failure_diagnostic(const hegel_failure_t *f);
...
 const char *hegel_failure_origin(const hegel_failure_t *f);
 
+/*
+ The failure's reproduce blob — a base64 string encoding the minimal
+ counterexample's choice sequence, suitable for deterministic replay via
+ `hegel_test_case_from_blob`. Returns NULL if `f` is NULL or the
+ engine produced no blob for this failure. The pointer is borrowed from the
+ parent `hegel_run_result_t` and stays valid until `hegel_run_free`.
+ */
+const char *hegel_failure_reproduce_blob(const hegel_failure_t *f);
```
DRMacIver: nit: I feel like this needs a non-verb name. `reproduction_blob` or something. Possibly we just need a better noun for these blobs in general. Right now it looks like this is the thing that reproduces the failure.

We don't have to resolve this in this PR though.

DRMacIver (approved):

hegel-c/include/hegel.h
echoumcp1: (summarized from convo elsewhere)

hegel_test_case_from_blob is a constructor returned to (and owned by) the caller, so it should have a free. There is no run when replaying from a blob. Regular test cases are owned by the run and destructed by the run (in Rust). The caller only gets a borrow of the test case from the run. Calling hegel_test_case_free on a test case owned by the run loudly fails with a clear error message explaining why that's not allowed.

PR #321, opened 2026-06-09
Improvements to stateful testing "variables" API

DRMacIver:
The first improvement is to rename Variables to Pool, because Variables sucks as a name (sorry).

The other improvement is to properly integrate it into the generators API rather than have its own draw methods. I always sortof intended those methods to be internal implementation details but then I just sortof... forgot to implement the public API corresponding to them.

You can now get two different generators off a Pool:

- `references()` yields `&T` without removing (the old `draw()`).
- `values()` yields `T`, removing it from the pool (the old `consume()`).

As well as being more unified with the rest of the ways we generate data and allowing you to use pool values in more places, this means that values drawn from pools in stateful tests will get printed properly (which is the actual thing that prompted me realising this needed doing).

This also removes the default Send + Sync bound from generators (because it was awkward for the pool generators) and applies it in the few places we need it. I think we need a better think through concurrency control on generators in general but I didn't want to make getting that right part of this PR.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,32 @@
+RELEASE_TYPE: minor
+
+This release reworks the stateful-testing value-pool API. The `Variables<T>`
+type is renamed to `Pool<T>` and the `variables()` constructor is renamed to
+`pool()`. The `draw()` and `consume()` methods are replaced by two generators
+you draw from with `tc.draw()`:
+
+- `pool.references()` returns a generator over `&T` — drawing from it yields a
+  reference to a value in the pool without removing it (the old `draw()`).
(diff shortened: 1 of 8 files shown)


Liam-DeVoe (approved):
API lgtm. My one doubt is the name `.values()` doesn't make it obvious drawing from this will consume a value from the pool. I don't have a better name suggestion right now

DRMacIver:
> API lgtm. My one doubt is the name `.values()` doesn't make it obvious drawing from this will consume a value from the pool. I don't have a better name suggestion right now

Yeah. It's not a great name.

It makes sense in a rust context, but maybe not in general.

Perhaps `reusable_values` and `consumed_values`? Awkward but explicit.

PR #326, opened 2026-06-10
Swarm testing

echoumcp1:
Port of https://github.com/hegeldev/hegel-core/pull/136


src/native/data_source.rs
```diff
@@ -208,6 +208,41 @@ impl DataSource for NativeDataSource {
         })
     }
 
+    fn new_state_machine(
+        &self,
+        rule_names: &[&str],
+        invariant_names: &[&str],
+    ) -> Result<i64, DataSourceError> {
```
DRMacIver, line 215: Why are our state machine IDs i64 rather than u64?

DRMacIver (commented):
I think this generally looks fine, but something I'm not sure about (that was present in the original as well, but feels worse in a C API)... Is there actually any good reason for us to pass the names of the rules and invariants? Would it not make more sense to just pass the number and have them passed back and forth as integer indices?

src/native/data_source.rs
echoumcp1: every engine-issued id in the DataSource trait (collection ids, pool ids, and variable ids all return Result<i64, …>) so I was trying to keep it consistent. That being said I made next_rule return u64 by accident so I will change that to i64. I think i64 is probably fine because Go uses signed ints for lengths and indices and Java lacks unsigned types entirely, so changing to u64 would require us to replace ints with longs.

DRMacIver: Hmm. That's fair. Let's keep it as i64. It's not like we're ever going to need that many anyway.

echoumcp1:
> I think this generally looks fine, but something I'm not sure about (that was present in the original as well, but feels worse in a C API)... Is there actually any good reason for us to pass the names of the rules and invariants? Would it not make more sense to just pass the number and have them passed back and forth as integer indices?

I can't really think of a good reason to pass the names of the rules either to be honest. I was probably overindexing on the Python implementation. Any thoughts @Liam-DeVoe

Liam-DeVoe:
I passed the invariant and rule names for observability

DRMacIver:
> I passed the invariant and rule names for observability

Hmm. That's a good reason.

And I guess we're  choosing the rules to run by index rather than name, so there's no big cost to this.

OK I'm sold.

DRMacIver (approved):

PR #327, opened 2026-06-10
Add primitive_boolean

echoumcp1:
Hegel libraries with stateful testing implement generation by drawing a step cap with `integers` and looping until the number of steps run successfully reaches the cap (see [stateful.rs](https://github.com/hegeldev/hegel-rust/blob/257bcd94a6013e22716c247cd3f46976024ff326/src/stateful.rs#L222)). In Hypothesis, after each step, we [draw from a weighted boolean generator with a vanishingly small probability of being true](https://github.com/HypothesisWorks/hypothesis/blob/fcc26c4c67eb9aaf4f153417f373f4c349128b54/hypothesis/src/hypothesis/stateful.py#L163) to decide when to break out of the generation loop, which allows the shrinker to stop generation at after any step. To allow Hegel libraries using the native backend to implement this, we add a `primitive_boolean` which allows control over probability and forcing.


hegel-c/tests/smoke.rs
```diff
@@ -743,6 +746,157 @@ fn libhegel_pool_primitives_draw_added_variables() {
...
+            let rc = (a.primitive_boolean)(tc, 1.0, false, false, &mut v);
+            assert_eq!(rc, 0, "p=1 draw failed: rc={}", rc);
+            assert!(v);
+
+            // An unforced fair draw; both outcomes must show up across the
+            // run. The draw can report STOP_TEST if the engine's choice
+            // budget is exhausted mid-shrink, so treat that the same way the
+            // other primitives do.
+            let rc = (a.primitive_boolean)(tc, 0.5, false, false, &mut v);
+            if rc == -1 {
```
DRMacIver: Why is this a -1 check rather than a named constant?

hegel-c/tests/smoke.rs
```diff
@@ -743,6 +746,157 @@ fn libhegel_pool_primitives_draw_added_variables() {
...
+            let tc = (a.next_test_case)(run);
+            if tc.is_null() {
+                break;
+            }
+            saw_test_case = true;
+
+            // NULL test-case handle is reported as HEGEL_E_INVALID_HANDLE.
+            let mut v = false;
+            let rc = (a.primitive_boolean)(ptr::null_mut(), 0.5, false, false, &mut v);
+            assert_eq!(rc, -4, "expected HEGEL_E_INVALID_HANDLE, got rc={}", rc);
```
DRMacIver: Same question.

DRMacIver (approved):
Question about the tests using raw integer literals rather than named constants, but once that's addressed don't need to rereview.

hegel-c/tests/smoke.rs
echoumcp1: The preexisting smoke tests also use explicit numbers in many places. I will change all to use constants.

DRMacIver: Whoops, sorry, that's my bad.

(It's always easier to notice problems like this in other people's code)

PR #335, opened 2026-06-18
Fix primitive_boolean not honoring small probabilities

echoumcp1:
Used in hegel-ocaml stateful testing. Approximation of 2^-16 as 2^-8 caused tests to stop generating before reaching stateful_step_count about 15 percentage points more often than in Hypothesis (~24% vs ~9%).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Fix `primitive_boolean` not honoring `p` (probability of drawing true) when it is very small.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,3 @@
 RELEASE_TYPE: patch
 
-Fix `primitive_boolean` not honoring `p` (probability of drawing true) when it is very small.
+Fixes stateful testing terminating step generation prematurely.
--- a/hegel-c/src/native/core/choices.rs
+++ b/hegel-c/src/native/core/choices.rs
@@ -137,10 +137,8 @@ impl BooleanChoice {
         false
     }
 
-    /// Misalignment fallback for a replayed boolean (see [`ChoiceKind::unit`]).
-    /// `false`, matching Hypothesis's simplest choice for booleans
     pub fn unit(&self) -> bool {
-        false
+        true
     }
 
     pub fn max_index(&self) -> crate::native::bignum::BigUint {
--- a/hegel-c/src/native/core/choices.rs
+++ b/hegel-c/src/native/core/choices.rs
@@ -137,8 +137,10 @@ impl BooleanChoice {
         false
     }
 
+    /// Misalignment fallback for a replayed boolean (see [`ChoiceKind::unit`]).
+    /// `false`, matching Hypothesis's simplest choice for booleans
     pub fn unit(&self) -> bool {
-        true
+        false
     }
 
     pub fn max_index(&self) -> crate::native::bignum::BigUint {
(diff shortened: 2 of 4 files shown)


hegel-c/src/native/core/choices.rs
```diff
@@ -137,8 +137,10 @@ impl BooleanChoice {
         false
     }
 
+    /// Misalignment fallback for a replayed boolean (see [`ChoiceKind::unit`]).
+    /// `false`, matching Hypothesis's simplest choice for booleans
     pub fn unit(&self) -> bool {
```
DRMacIver, line 140: This is incorrect. The unit is the *second* simplest element (It's the local equivalent of 1, hence "unit").

hegel-c/src/native/core/state.rs
```diff
@@ -545,6 +545,24 @@ pub(crate) fn weighted_boolean_sample(p: f64, rng: &mut EngineRng) -> bool {
...
+/// floor (which would turn e.g. a stateful stop signal's `p = 2^-16` into
+/// `1/256`).
+///
+/// Spends 8 bytes of entropy rather than one, so it is reserved for draws
+/// whose probability needs the precision (routed via
+/// [`NativeTestCase::weighted_precise`]); ordinary booleans keep the one-byte
+/// sampler. Callers must pass `0.0 < p < 1.0`; boundaries are forced without
+/// entropy by `weighted_precise`.
+pub(crate) fn weighted_boolean_sample_precise(p: f64, rng: &mut EngineRng) -> bool {
+    let mut buf = [0u8; 8];
```
DRMacIver: I think there's no benefit to doing this over using https://docs.rs/rand/latest/rand/trait.RngExt.html#method.random_bool which uses essentially the same mechanism, right?

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Fix `primitive_boolean` not honoring `p` (probability of drawing true) when it is very small.
```
DRMacIver: `primitive_boolean` isn't part of the public API, so this should reference the concrete problem this was fixing (stateful testing) rather than this.

DRMacIver (changes requested):
As well as the concrete fixes, meta-comment: I think what's happened here is a misunderstanding on Claude's part (and also mine actually, this changed since I wrote the code and I got confused by the new code), which is that in Hypothesis we *do* quantize booleans this way, but only for the bytestring backend (which we use for fuzzing).

I think switching to using a full  set of bytes for weighted booleans is OK as an intermediate fix, but it's bad for the URandom backend, where it means in the Antithesis case we're consuming way more entropy from the fuzzer than is ideal.

There's an entropy-parsimonious way we can do this that we can talk about implementing later but don't have to do now. Basically it works like this:

```python
import math

def draw_weighted(rng, p):
     while True:
         quantized_true = math.floor(p * 256)
         quantized_false = math.floor((1 - p) * 256)
         i = rng.getrandbits(8)
         if i <= quantized_true:
             return True
         elif i <= quantized_true + quantized_false:
             return False
         assert i == 255
         p = 256 * p - quantized_true
         assert 0 <= p <= 1         
```

(disclaimer: Haven't checked this code is right, it's from memory and lightly adapted to the byte case, I've probably made a mistake)

The idea being roughly that you reserve the last of your 256 values as a correction term for the error you get by quantizing.

We don't need to do this as part of this PR, this is just FYI. Hypothesis also gets this wrong, but it's somewhat less of a big deal for Hypothesis whether Antithesis integration works smoothly.

hegel-c/src/native/core/state.rs
echoumcp1: yeah, probably Claude copying weighted_boolean too closely

echoumcp1: Double checked with a dummy stateful test and it made no difference in the percentage of tests reaching the full step count.

DRMacIver (approved):

PR #337, opened 2026-06-18
Shrinker improvements

echoumcp1:
Reduced test body executions during shrinking by 6-9x in a stateful test mimicking one from the Aria team. Seems like an improvement in general?

### DIFF
--- a/hegel-c/src/native/shrinker/mod.rs
+++ b/hegel-c/src/native/shrinker/mod.rs
@@ -243,14 +243,8 @@ impl<'a> Shrinker<'a> {
         if sort_key(nodes) == sort_key(&self.current_nodes) {
             return Ok(true);
         }
-        // ported from Hypothesis's `cached_test_function` (shrinker.py):
-        // compare the candidate against the current target over
-        // the target's length window (truncating a candidate longer than the
-        // target — extra trailing nodes can't make it shortlex-smaller). If the
-        // truncated candidate is already shortlex >= the current target, no
-        // realised run from it can improve on the target, so reject it without
-        // running the test. Deletion candidates are shorter, hence
-        // shortlex-smaller, and are NOT rejected here — they still run.
+        // If the truncated candidate is already shortlex >= the current target, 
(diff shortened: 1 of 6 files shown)


hegel-c/src/native/shrinker/mod.rs
```diff
@@ -243,16 +243,44 @@ impl<'a> Shrinker<'a> {
...
+        // draws re-recorded as forced throughout a stateful run), which is the
+        // single biggest blocker to deleting whole steps. For length-changing
+        // candidates the realised run re-derives forced-ness anyway, and
+        // `accept_improvement` only commits a strictly-smaller result, so
+        // skipping the guard here is safe.
         for (i, candidate) in nodes
             .iter()
             .enumerate()
             .take(nodes.len().min(self.current_nodes.len()))
+            .filter(|_| nodes.len() == self.current_nodes.len())
```
DRMacIver: This seems a *very* weird way to write what is in fact an if statement that short-circuits this loop body? I take it Claude didn't want to indent a bunch of code?

echoumcp1: The OCaml side of me probably would have also written it this way. But I could change it to an if statement if you prefer.

DRMacIver: Pretty strong preference for an if statement here yeah.

hegel-c/src/native/shrinker/mod.rs
```diff
@@ -243,16 +243,44 @@ impl<'a> Shrinker<'a> {
         if sort_key(nodes) == sort_key(&self.current_nodes) {
             return Ok(true);
         }
+        // ported from Hypothesis's `cached_test_function` (shrinker.py):
```
DRMacIver: Not necessarily to change here, but I think we should have a think about how to get the Claude comments under control. They're really over-verbose and I'm not sure how helpful they are. This isn't a *bad* comment, but I don't know that it adds much.

DRMacIver (approved):

PR #338, opened 2026-06-18
Return `std::net::IpAddr` instead of `String` in `ip_addresses` generator

### LINKED ISSUE #234: Return `std::net::IpAddr` in the `ip_addresses` generator
Rust has a stdlib representation of IpAddr. We should return that instead of a string.

owenbrooks:
I removed the existing string-based tests since they no longer seem relevant. We pass through `std::net:IpAddr`'s `FromStr` that enforces those properties internally. The basic generation test is still covered by `check_can_generate_examples` in `test_ip_addresses_kwargs`.

Please let me know if this is what you were envisioning.

Should I implement DefaultGenerator for IpAddr (and V4/V6)?

Resolves #234


DRMacIver:
Thanks! This looks good.

And yeah it would be nice to add default generators for the types if we're doing this. If you wouldn't mind adding that to the PR as well that would be great, thanks.

DRMacIver:
BTW this will require a RELEASE.md (which should be a minor release) for the build to pass, so that it can get released when it's merged to main. Take a look at RELEASE-sample.md for what that looks like.

owenbrooks:
Thanks for the hint on RELEASE.md.

I added `impl DefaultGenerator for IpAddr`. Unless I'm missing something, it doesn't seem straightforward to do the same for Ipv4Addr and Ipv6Addr without adding specialised V4/V6 Generators in addition to IpAddressGenerator.

I can go ahead and do that if you want, or we could skip DefaultGenerator for V4/V6 variants, unless you have another idea.

DRMacIver:
Hmm.  Adding specialised V4/V6 Generators seems sortof reasonable. We could change it so that the `v4` and `v6` methods on IpAddressGenerator return an IpV{4,6}AddressGenerator type instead of Self.

Up to you if you want to do this though. What you've got here seems a strict improvement on what we've got so far and I'm happy to merge as is if you'd prefer not to.

owenbrooks:
I've created these specialised Ipv{4,6}AddressGenerator types, returned them from the v4() and v6() methods, and implemented DefaultGenerator for Ipv4Addr and Ipv6Addr.

Liam-DeVoe (approved):
Perfect, thanks!

(pending one small docs issue in CI, which you're welcome to fix if you want, or I'll get to otherwise)

PR #340, opened 2026-06-18
Expose number of choices in test case 

echoumcp1:
Implement this [check](https://github.com/HypothesisWorks/hypothesis/blob/5f3f6ae6599e5d9ebaeeebaf0df6f9fc011e308f/hypothesis/src/hypothesis/stateful.py#L156) in stateful testing.

### DIFF
--- a/hegel-c/src/lib.rs
+++ b/hegel-c/src/lib.rs
@@ -1340,16 +1340,17 @@ pub unsafe extern "C" fn hegel_start_span(
 /// choice buffer overruns.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn hegel_test_case_length(
+    ctx: *mut HegelContext,
     tc: *mut HegelTestCase,
     out_count: *mut usize,
-) -> c_int {
-    clear_last_error();
+) -> hegel_result_t {
+    clear_last_error(ctx);
     let tc = match unsafe { tc_mut(tc) } {
         Ok(t) => t,
         Err(rc) => return rc,
     };
     if out_count.is_null() {
-        set_last_error("hegel_test_case_length: out parameter is null");
+        set_last_error(ctx, "hegel_test_case_length: out parameter is null");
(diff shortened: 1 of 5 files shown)


hegel-c/src/lib.rs
```diff
@@ -1333,6 +1333,30 @@ pub unsafe extern "C" fn hegel_start_span(
     }
 }
 
+/// Number of choices drawn so far in this test case (Hypothesis's
+/// `ConjectureData.length`). Writes the count into `*out_count` and returns
```
Liam-DeVoe, line 1337: fyi `ConjectureData.length` is the entropy of the choices, not the number of choices. iirc `length` is defined as the size in bytes of the serialized choice

echoumcp1: Wonder if Claude made the choice to make it [number of choices](https://github.com/hegeldev/hegel-rust/blob/3c689e077d169fe74a40697bfff3ea0c25af5432/hegel-c/src/native/core/mod.rs#L22) in the Rust port? Analogous part in [Hypothesis](https://github.com/HypothesisWorks/hypothesis/blob/720e19cb9ae313427283e5fb560c7c78ddb2936b/hypothesis/src/hypothesis/internal/conjecture/engine.py#L99) says no defined semantics

Liam-DeVoe: oh yeah looks like claude has made the wrong decision there. cc @DRMacIver. This means hegel-native is going to take much longer to overrun test cases than hypothesis, which could be contributing to some of our observed slowdowns.

That comment about semantics is more for end users who want to monkeypatch that constant. The choice of unit does have substantial implications for the performance of the engine.

DRMacIver:
I think independently of whether choices is the right metric (I think it's probably fine), this is rendered redundant by #349 right?

echoumcp1:
> I think independently of whether choices is the right metric (I think it's probably fine), this is rendered redundant by #349 right?

yes

PR #348, opened 2026-06-24
Expand PathBuf DefaultGenerator to generate non-UTF-8 paths

### LINKED ISSUE #216: Consider generating non-utf8 `PathBuf` instances in our `DefaultGenerator` implementation
See https://github.com/hegeldev/hegel-rust/pull/215#issuecomment-4319778497.

owenbrooks:
On Windows and Unix, this uses the methods provided in the respective `OsStringExt` traits, namely `from_wide` and `from_bytes`. On other platforms it falls back to UTF-8 Strings. 

These methods are discussed in https://doc.rust-lang.org/std/ffi/index.html#conversions

A quick check of other platforms in `std::os::ffi` source for similar `OsStringExt` impls showed only 'Motor OS', but it assumes OsStrings are UTF-8 so it is covered by the default case.

Fixes #216


DRMacIver:
Thanks!

I'd like to have some tests for this - even something as simple as asserting that we can find a path that isn't valid UTF-8. We've got the `find_any` helper for testing generation.

Also I'm not sure I believe that the non-unix non-windows case works at all in Hegel right now. I don't mind supporting it here, but it's not tested and as a result I don't expect it to work.

sunshowers:
This is relevant to my interests since I maintain [camino](https://docs.rs/camino), which does assert that paths are UTF-8. camino already has a proptest feature, and I'm very interested in adding hegel support as well, but I'm a bit hesitant to add support for it behind a feature while hegel is in 0.x. I'm wondering if there's any appetite in moving just some kind of generator trait into a stable 1.x crate.

DRMacIver:
> I'm wondering if there's any appetite in moving just some kind of generator trait into a stable 1.x crate.

I think not yet I'm afraid. There are a bunch of design questions around the generator trait that we're still legitimately unsure about, and I don't want to prematurely stabilise it.

(The big one is that we still have some questions about the right way to do printing)

PR #352, opened 2026-06-27
fix a couple of Tree Borrows violations detected by Miri

sunshowers:
While trying to hook up Hegel with Miri, I found a couple of Tree Borrows mutable aliasing violations in this code.

```rust
            let tc = Box::new(HegelTestCase {
                ds,
                completed: false,
                last_value: Vec::new(),
                ack: Some(ack),
            });
            let case = (&*tc) as *const HegelTestCase as *mut HegelTestCase;
            run.current_tc = Some(tc);
```

There are two violations here:

1. The `let case` line turned a `&T` into a `*mut T`. This by itself is fine, but actually dereferencing the pointer to mutate it (as callers do) would effectively mean that one was writing to data via a `&T` reference, which is forbidden in Rust (unless `UnsafeCell` is involved, which it isn't here).
2. We were storing a `Box<_>`, which asserts both ownership and a noalias restriction. The first one is correct, but the second one is not: we were returning a mutable pointer to the test case via the `*out_test_case` pointer, and then other code could mutate the underlying memory via that pointer.

Fix both of these by storing a logically-owned pointer instead (`*mut T` does not assert noalias), and manually managing memory on drop.

I didn't set up an actual Miri run here -- you're probably best positioned to do that.

### DIFF
--- a/hegel-c/src/lib.rs
+++ b/hegel-c/src/lib.rs
@@ -453,7 +453,14 @@ pub struct HegelRun {
     worker: Option<JoinHandle<()>>,
     from_worker: mpsc::Receiver<WorkerMessage>,
     abort: Arc<AtomicBool>,
-    current_tc: Option<Box<HegelTestCase>>,
+    // The current test case.
+    //
+    // This is a logically owned pointer, and would be
+    // `Option<Box<HegelTestCase>>` but for the fact that `Box` also asserts
+    // noalias (i.e., that there are no mutable references that aren't derived
+    // from `current_tc`). Since the caller mutates the current test case
+    // through a different pointer, we use a raw pointer instead.
+    current_tc: Option<*mut HegelTestCase>,
     result: Option<HegelRunResult>,
     /// Set once a TestRunDone (or worker-died Err) has been observed on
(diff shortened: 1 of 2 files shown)


hegel-c/src/lib.rs
```diff
@@ -1145,7 +1160,11 @@ pub unsafe extern "C" fn hegel_run_free(
 
     run.abort.store(true, Ordering::Release);
 
-    if let Some(mut tc) = run.current_tc.take() {
+    if let Some(tc) = run.current_tc.take() {
+        // SAFETY: `run.current_tc` only ever holds a live pointer created by
+        // `Box::into_raw`, and the caller is expected to not dereference this
+        // pointer once it has been freed here.
+        let mut tc = unsafe { Box::from_raw(tc) };
```
sunshowers, line 1167: I think I got the safety comments right here -- let me know if not. This stuff is always hairy.

DRMacIver: I'm honestly very bad at reasoning about unsafe rust. It looks right to me, but I'm also more confident in your read than mine!

DRMacIver:
Thanks for this! Running Miri in CI sounds like a very good idea. I'll get that set up.

PR #353, opened 2026-06-27
Don't copy doc comments to state machine rules

brownben:
Currently, all attributes applied to functions in a state machine are also applied to the list of rules.

Doc comments are desugared into attributes and thus are also copied to the list of rules. This leads to the `unused_doc_comments` lint firing,  as the doc comment should only be on the function, not in the list.

Only pass through `cfg` attributes, rather than all attributes. As the [other attributes listed in the Rust reference](https://doc.rust-lang.org/reference/attributes.html#built-in-attributes-index) all look like they should only be applied to the function rather than references to the function.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This release prevents the `unused_doc_comments` lint from firing when using doc comments on stateful test `#[rule]`s or `#[invariant]`'s.
--- a/hegel-macros/src/stateful.rs
+++ b/hegel-macros/src/stateful.rs
@@ -42,7 +42,12 @@ pub fn expand_state_machine(mut block: ItemImpl) -> TokenStream {
 
             let info = || MethodInfo {
                 name: method.sig.ident.clone(),
-                attrs: method.attrs.clone(),
+                attrs: method
+                    .attrs
+                    .iter()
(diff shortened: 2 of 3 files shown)


DRMacIver:
Hmm. The doc comments thing is a good catch and we should definitely fix that, thanks for the report, but I don't think this is the right fix. The original motivation wasn't cfg at all, but that there might be other attribute-style macros on the rule, and we wanted to pass those along.

Liam-DeVoe:
I actually agree with Ben's reasoning here. The crux is that the user has defined attributes on a `#[rule]`, which we are then forwarding to a specific `Rule` instance in a `vec`. The latter is an *expression* ("statement"?), which has different semantics for what rust attributes are valid. Rust has this to say:

>   ▎ The attributes that have meaning on a statement are cfg, and the lint check attributes.

Which implies we want to forward only the following attributes:

```rust
            const FORWARDED: [&str; 7] = [
                "cfg", "cfg_attr", "allow", "expect", "warn", "deny", "forbid",
            ];
```

For example on master we currently have a bug where we forward the following rustfmt attr, which errors:

```rust
        #[rustfmt::skip]
        #[rule]
        fn unformatted(&mut self, _tc: TestCase) {
            self.count += 1;
        }
```

I will put up a PR.

Liam-DeVoe:
Thanks @brownben!

PR #359, opened 2026-06-29
Add cloneable test-case handles to the libhegel C API

DRMacIver:
This lays the foundation for getting concurrent generation working well. It doesn't actually make concurrent generation work well, it just gives C API level support for the semantics we'd already implemented in hegel-rust and makes incorrect use more prominently breaking.

Basically it adds two new rules:

* test cases cannot be used from multiple threads concurrently and will error if you try (NB: Can be used from multiple threads sequentially just fine). You can clone test cases and it is safe to use clones concurrently.
* All test cases now must be freed by the caller, regardless of where they come from. This should hopefully make GC integration much easier and is the only thing that isn't a complete nightmare once you start cloning test cases and want proper concurrency support, but it's a biggish breaking change.

<details>
<summary>AI-generated implementation & test summary</summary>

Exposes the underlying API for concurrent data generation. The hegel-rust `TestCase` already encodes the target shape (`Clone + Send + !Sync`); this brings the libhegel C ABI up to match and rewires the frontend onto it. The engine model is unchanged — concurrent draws across clones are memory-safe but non-deterministic for now; making them robust is future work.

**libhegel C ABI (`hegel-c/src/lib.rs`)**
- New `hegel_test_case_clone(ctx, tc, out)` returns a handle sharing the underlying test case (the "family") — same data source, shared completion.
- `HegelTestCase` split into a shared `FamilyShared { ds: Arc<dyn DataSource>, completed: AtomicBool, ack, clones }` and per-instance `Mutex<LocalState>` (the `last_value` buffer) + `is_root`.
- Per-instance lock via `try_lock`: a single handle used from two threads returns the new `HEGEL_E_CONCURRENT_USE`; two clones (independent locks) may draw concurrently.
- `hegel_mark_complete` on any handle marks the whole family complete via a `compare_exchange` gate (engine report + run ack fire exactly once).
- Only the root may be freed; freeing it cascades to every clone. Freeing a clone returns the new `HEGEL_E_NOT_ROOT`. Run-owned roots still return `HEGEL_E_INVALID_HANDLE` (freed by the run).
- All handle allocations go through a `Send + Sync`-bounded constructor, making the thread-safety contract a compile-time check.

**hegel-rust frontend**
- `TestCase` holds a per-instance non-owning handle; `clone` forks a fresh libhegel handle via `hegel_test_case_clone`, `child` shares the same handle. Draw-name bookkeeping stays behind a blocking mutex.
- `run_test_case` runs the body on the root handle and reports the outcome via the retained root, so a libhegel clone is created only on a genuine `tc.clone()`, never once per case.

**Tests** — new C-ABI tests (clone null/happy/clone-of-clone/after-complete, `NOT_ROOT`, shared completion, both free cascades, two clones drawing concurrently), a deterministic `HEGEL_E_CONCURRENT_USE` embedded test, and hegel-rust tests for `clone_handle` and cross-thread `TestCase::clone`.

**Verification** — `just check` (fmt/clippy/all sub-lints/tests), 100% line coverage (nocov ratchet unchanged at 322), `just c-test` (smoke + C examples), and `just check-miri` (no UB) all pass; the cbindgen header is regenerated and in sync.

</details>


DRMacIver:
@lmb Your feelings on this PR would be very much appreciated, as it's probably going to impact hegel-go a lot.

(Asking more about semantics than implementation right now, still reviewing the implementation myself)

hegel-c/include/hegel.h
```diff
@@ -22,23 +22,32 @@
...
+ * matching free — every handle from these:
  *
  *     hegel_context_new          ->  hegel_context_free
  *     hegel_settings_new         ->  hegel_settings_free
  *     hegel_run_start            ->  hegel_run_free
  *     hegel_test_case_from_blob  ->  hegel_test_case_free
+ *     hegel_next_test_case       ->  hegel_test_case_free
+ *     hegel_test_case_clone      ->  hegel_test_case_free
+ *
+ * Every test-case handle you receive — whether from hegel_test_case_from_blob,
```
lmb, line 37: Yes, this is nicer.

hegel-c/include/hegel.h
```diff
@@ -427,16 +444,26 @@ typedef struct hegel_run_result_t hegel_run_result_t;
...
- `hegel_test_case_free`.
+ One in-flight test-case handle handed to the caller by
+ `hegel_next_test_case`, `hegel_test_case_from_blob`, or
+ `hegel_test_case_clone`. The caller drives it with the per-test-case
+ primitives (`hegel_generate`, `hegel_start_span` / `hegel_stop_span`,
+ `hegel_target`, the collection primitives) and concludes it with
+ `hegel_mark_complete`.
+
+ A single handle must be driven by at most one thread at a time: each
+ primitive `try_lock`s the handle's own `local`, returning
```
lmb: Nit: implementation detail.

hegel-c/include/hegel.h
```diff
@@ -22,23 +22,32 @@
...
  * Every *other* pointer libhegel hands back is borrowed: libhegel still owns
  * it, you must not free it, and it is valid only until a point that the
  * function documents. Two cases are easy to trip over:
  *
- *   - The hegel_test_case_t* from hegel_next_test_case is borrowed from the
- *     run and freed by hegel_run_free. Do NOT pass it to hegel_test_case_free
- *     (that is only for a test case you made with hegel_test_case_from_blob).
- *     Likewise the hegel_run_result_t* and hegel_failure_t* you read from a
- *     run live until hegel_run_free.
+ *   - The hegel_run_result_t* and hegel_failure_t* you read from a run are
```
lmb: Could we also make this something that has to be freed? Alternatively, why are they separate objects if they are always tied to the lifetime of the run?

hegel-c/include/hegel.h
```diff
@@ -678,27 +708,58 @@ hegel_result_t hegel_run_free(hegel_context_t *ctx, hegel_run_t *run);
...
+
+ The clone is a *view onto the same test case*, not an independent one: it
+ draws from the same data source, and `hegel_mark_complete` on any handle in
+ the family marks them all complete. Clones exist so a test case can be
+ driven from several threads — each handle has its own lock, so two clones
+ may draw concurrently, whereas using a *single* handle from two threads
+ returns `HEGEL_E_CONCURRENT_USE`. (Concurrent draws across clones are
+ currently non-deterministic; making them robust is future work.)
+
+ Cloning is allowed on a clone (the result shares the same family) and
```
lmb, line 778: Introduces new family concept. Should that be at the top?

hegel-c/include/hegel.h
```diff
@@ -678,27 +708,58 @@ hegel_result_t hegel_run_free(hegel_context_t *ctx, hegel_run_t *run);
...
  */
 hegel_result_t hegel_test_case_free(hegel_context_t *ctx, hegel_test_case_t *tc);
 
+/*
+ Clone a test-case handle, writing a new handle that shares the same
+ underlying test case into `*out_test_case`.
+
+ The clone is a *view onto the same test case*, not an independent one: it
+ draws from the same data source, and `hegel_mark_complete` on any handle in
+ the family marks them all complete. Clones exist so a test case can be
```
lmb, line 772: This leaves out what happens to other test cases in the same family when one is marked complete. Do I as a library author have to do anything else? It would be very very nice if I didn't have to worry about this. AKA first mark_complete "cancels" other test_runs, but calling mark_complete on other test_runs in the same family is OK. (otherwise i somehow have to coordinate only calling mark_complete from one thread).

DRMacIver: Yeah I've been trying to decide what the right semantics for this are and I'm a bit unsure. They're super awkward regardless of what we do.

I'm somewhat inclined to make `mark_complete` an error on clones, but I think that makes it difficult to implement e.g. concurrent stateful testing well.

In particular we have to worry about e.g. what happens when one thread marks complete interesting, and another invalid.

I'm kindof inclined to make it have "first caller wins" semantics. It doesn't protect the user from races, but it's the most flexible option.

hegel-c/include/hegel.h
DRMacIver: Hmm I'm not sure "family" is a concept we need to expose to the user. Though maybe it is because of the mark_complete issues discussed below.

(Sorry these docs are still totally AI generated and I need to make them better)

hegel-c/include/hegel.h
DRMacIver: Yeah. I already knew what we had before was gross but was a bit unsure how to fix it optimally until the cloning work forced me to think about it harder, but I think this is clearly what we should have been doing all along.

hegel-c/include/hegel.h
DRMacIver: Hmm. Good question. Will look into it.

hegel-c/include/hegel.h
lmb: you could make it error, but that implies we as consumers need to do something with that error. my preference would be "first caller wins, without an error from losing mark_complete". followed by "losing mark_complete returns a dedicated error" so i can ignore it in the caller (because i don't know what to do with it).

DRMacIver: I think first caller wins without an error is probably the best starting point, though it's not the ideal long-term solution.

PR #364, opened 2026-07-07
hegel-c: allow setting output per hegel_context_t

### LINKED ISSUE #355: hegel-c: allow setting output per hegel_context_t
From what I can tell hegel-c currently outputs to stderr / stdout. We expose a knob to change verbosity, but do not allow to redirect where things are written. I'd like a knob per `context` which allows me to control where things are written.

This could either take a callback `error_t (uint8_t*, size_t)` or a file descriptor, my preference being a callback. File descriptors are probably a lot easier to integrate into hegel-rust, but they make redirection inside a single process a lot more expensive.

As an example, in Go I want to redirect all output into a `testing.T` which gives me a `func([]byte) error`. Doing this via a file descriptor involves:

- Create a pipe
- Pass the write-end to hegel-c
- Ensure I close the write-end
- Spawn a goroutine which reads from read-end and invokes the `func`
- Interrupt the goroutine if the test ends prematurely (by closing the read-end)
- Avoiding a goroutine leak if the test exits normally

The callback has the drawback of having more complicated FFI story because now we have C / Rust code invoking Go code.

DRMacIver:
Closes #355.

<details>

## C ABI

- `hegel_context_t` grows an output destination — a `callback` function pointer plus an arbitrary `user_data` pointer — set with the new `hegel_context_set_output(ctx, callback, user_data)`. Passing a NULL callback resets to the default stderr output.
- The callback is invoked once per line of engine output (NUL-terminated UTF-8 plus explicit length, no trailing newline), with `user_data` passed through verbatim — so Go can register one library-wide function and resolve `user_data` to a `testing.T` handle.
- `hegel_run_start` and `hegel_test_case_from_blob` snapshot the context's destination at creation time; the engine's worker thread emits through the snapshot for the life of the run, so re-pointing or resetting the context later never races an in-flight run.
- Internally the engine's `Settings` carry an `Output` sink (default stderr); every `eprintln!` in `test_runner.rs` and the blob-replay path in `embed.rs` goes through it.

## Rust frontend

No new public API. The crate-internal `with_output_override` now captures *all* of a run's output, not just the frontend's draw/note lines:

- A run resolves the override **once, at start**, into a `RunOutput` it carries for its lifetime — never reading the thread-local at emit time, which would misroute output produced off the starting thread (the engine's worker, test-case clones driven on other threads).
- Engine output is wired in by installing the run's sink as the context callback around `hegel_run_start` / `hegel_test_case_from_blob` (trampoline in `ffi.rs`; the boxed sink is owned by the `RunHandle` and freed after `hegel_run_free` joins the worker).
- The final failure diagnostics, reproducer line, and multi-failure headline now also flow through the run's sink (unchanged on stderr when no override is installed).
- Explicit test cases emit through the same mechanism.

## Testing

- In-process C-ABI tests: callback receives generation/shrink debug lines with `user_data` passed through, NULL callback resets to stderr, NULL ctx rejected, run-start snapshot semantics, blob-replay trace routing.
- Frontend tests: `with_output_override` captures engine progress/shrink/replay lines, failure diagnostic + reproducer, multi-failure headline, and notes from a clone on another thread.
- Miri: the C-ABI full-run test installs the callback (cross-thread raw-pointer path), and a new `test_miri` case drives the frontend trampoline; `just check-miri` passes.
- New `hegel-c/examples/custom_output.c` exercises the feature end-to-end through the generated header under `just c-test` (shared + static).
- `just check`, `just c-test`, and `just check-coverage` (0 uncovered lines, nocov count unchanged) all pass.

</details>

### DIFF
--- a/hegel-c/include/hegel.h
+++ b/hegel-c/include/hegel.h
@@ -476,6 +476,9 @@ typedef enum {
  threads is a data race and unsupported. Passing `NULL` wherever a context
  is accepted is allowed and simply opts out of error messages: the call
  still returns its usual error code, there is just nothing to read back.
+
+ Besides error reporting, a context carries the output destination for the
+ runs and test cases created with it — see `hegel_context_set_output`.
  */
 typedef struct hegel_context_t hegel_context_t;
 
@@ -574,6 +577,16 @@ typedef struct hegel_string_generator_t hegel_string_generator_t;
 typedef struct hegel_test_case_t hegel_test_case_t;
 
 /*
+ Per-line output callback, installed on a context with
+ `hegel_context_set_output` (see there for the full contract). `user_data`
+ is the pointer registered alongside the callback; `line` is one line of
+ engine output, NUL-terminated UTF-8 of `len` bytes (not counting the
+ terminator) without a trailing newline, valid only for the duration of
+ the call.
+ */
+typedef void (*hegel_output_callback_t)(void *user_data, const char *line, size_t len);
+
+/*
  An engine-allocated byte buffer returned by `hegel_generate_bytes`.
 
  The caller owns the buffer and must release it with
@@ -642,6 +655,37 @@ extern "C" {
  Never returns NULL. Must be paired with a `hegel_context_free` call.
  */
 hegel_context_t *hegel_context_new(void);
+
+/*
+ Redirect engine-emitted output (verbose / debug progress traces, warnings)
+ for the runs and test cases subsequently created with `ctx`, instead of
+ writing it to stderr.
+
+ `callback` is invoked once per line of output; `user_data` is passed
+ through to every invocation verbatim, for the caller to resolve to
+ whatever sink the output should reach (say, a Go `testing.T` handle). The
+ `line` argument is NUL-terminated UTF-8 of `len` bytes (not counting the
+ terminator), without a trailing newline; the buffer is owned by libhegel
+ and valid only for the duration of the call — copy it to keep it. Passing
+ a NULL `callback` resets the context to the default stderr output
+ (`user_data` is ignored).
+
+ The destination is snapshotted when a run (`hegel_run_start`) or blob
+ replay (`hegel_test_case_from_blob`) is created, so changing it later
+ only affects subsequently created ones. The engine emits output from a
+ worker thread inside libhegel, so the callback must be safe to invoke
+ from a thread other than the one that registered it, and it — along with
+ whatever `user_data` points to — must stay valid until every run started
+ while it was installed has been freed with `hegel_run_free`.
+
+ This sets only the *destination*; how much output the engine emits is
+ controlled by `hegel_settings_set_verbosity`. Returns
+ `HEGEL_E_INVALID_HANDLE` for a NULL `ctx` (there is no context to store
+ the callback on).
+ */
+hegel_result_t hegel_context_set_output(hegel_context_t *ctx,
+                                        hegel_output_callback_t callback,
+                                        void *user_data);
 
 /*
  Free a context previously returned by `hegel_context_new`. Safe to call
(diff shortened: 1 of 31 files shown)


DRMacIver:
@lmb Does this have the semantics you want for go? I *believe* that this should do the right thing (you pass an opaque go Handle as the user_data and a single global function as the actual callback)

One oddity worth mentioning that I'm pretty sure is the correct choice but you might have thoughts and feelings on: Setting the callback is in the context, but the callback is propagated through test cases and runs, and if you were to set the callback after creating one of those it would be ignored until you create the next one.

The reason for this is because otherwise in implementations where you want to have the context be a thread local you run into problems with test cases being passed to other threads.

hegel-c/include/hegel.h
```diff
@@ -560,6 +563,16 @@ typedef struct hegel_string_generator_t hegel_string_generator_t;
...
 
+/*
+ Per-line output callback, installed on a context with
+ `hegel_context_set_output` (see there for the full contract). `user_data`
+ is the pointer registered alongside the callback; `line` is one line of
+ engine output, NUL-terminated UTF-8 of `len` bytes (not counting the
+ terminator) without a trailing newline, valid only for the duration of
+ the call.
+ */
+typedef void (*hegel_output_callback_t)(void *user_data, const char *line, size_t len);
```
lmb, line 587: Shape looks good!

hegel-c/include/hegel.h
```diff
@@ -628,6 +641,37 @@ extern "C" {
...
+ `callback` is invoked once per line of output; `user_data` is passed
+ through to every invocation verbatim, for the caller to resolve to
+ whatever sink the output should reach (say, a Go `testing.T` handle). The
+ `line` argument is NUL-terminated UTF-8 of `len` bytes (not counting the
+ terminator), without a trailing newline; the buffer is owned by libhegel
+ and valid only for the duration of the call — copy it to keep it. Passing
+ a NULL `callback` resets the context to the default stderr output
+ (`user_data` is ignored).
+
+ The destination is snapshotted when a run (`hegel_run_start`) or blob
```
lmb: I think this is weird, especially because we pass context into every single function. I guess there are behind the scenes reasons why this happens?

Maybe better to just have separate setters for a run or testcase from blob.

DRMacIver: The problem is that we can't guarantee that every function call gets the same context, and we want the callback to be stable between a test case and its clones.

DRMacIver: Being able to set it on the run or blob is an option. I disliked it when I tried it, and I think doing it causes us to grow an unpleasantly large API surface there, but I can see the argument for it.

DRMacIver: How about this: Test cases and runs inherit output sinks as in these semantics, but if you pass a different one in their context argument that is used instead.

I'm not 100% sure I love this semantics, but I think it probably results in the least surprising behaviour.

lmb: > and we want the callback to be stable between a test case and its clones.

Why? Feels like that is up to the library author to decide.

DRMacIver: Because the expected usage pattern is that contexts are thread local, and if you start generating from a new thread and suddenly your notes are going to the wrong place that would be terrible.

lmb: I'm a bit confused. I'd provide my own context (which is local to my unit of concurrency) in every call. How can output suddenly go to the wrong place?

DRMacIver: Because your test is not local to your unit of concurrency.

DRMacIver: Suppose I am a user of Hegel. I clone my test case, pass it to another thread, and call note on the test case in that thread. What gives that note call the right output target?

lmb: I think we're talking past each other :sweat_smile: Lets discuss in person?

PR #365, opened 2026-07-07
hegel-c: patch some validation gaps

lmb:


hegel-c/src/native/draws/mod.rs
```diff
@@ -73,6 +73,16 @@ pub fn generate_float(ntc: &mut NativeTestCase, spec: &FloatSpec) -> Result<f64,
...
         ));
     }
+    if spec.allow_nan && (spec.min_value.is_finite() || spec.max_value.is_finite()) {
+        return Err(EngineError::InvalidArgument(
+            "Cannot have allow_nan=true with min_value or max_value".to_string(),
+        ));
+    }
+    if spec.allow_infinity && spec.min_value.is_finite() && spec.max_value.is_finite() {
+        return Err(EngineError::InvalidArgument(
+            "Cannot have allow_infinity=true with both min_value and max_value".to_string(),
```
DRMacIver, line 83: These error messages are technically wrong, right? They should say "finite min_value and max_value" (you can set an explicitly infinite value).

DRMacIver (approved):
One minor comment but otherwise LGTM

hegel-c/src/native/draws/mod.rs
lmb: FWIW Claude just copied this from elsewhere:

```
$ git switch main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.

[~/dev/hegel-rust]$ rg -F "both min_value and max_value"
tests/test_validation.rs
277:            "allow_infinity=true with both min_value and max_value",

src/generators/numeric.rs
322:            invalid_argument!("Cannot have allow_infinity=true with both min_value and max_value");

tests/test_standard_generators.rs
335:            "allow_infinity=true with both min_value and max_value",
```

I'll leave it and you can follow up if you feel strongly about it?

DRMacIver: huh fair enough. I'll try to do an audit pass over error messages at some point, feel free to leave as is.

PR #367, opened 2026-07-08
Result of asking Fable to do a big review and fixup on the codebase

DRMacIver:


src/cli.rs
```diff
@@ -201,7 +201,7 @@ fn usage() -> String {
     s.push_str("  --verbosity <LEVEL>                  quiet | normal | verbose | debug\n");
     s.push_str("  --derandomize <true|false>           Use a deterministic derived seed\n");
     s.push_str(
-        "  --database <PATH|disabled>           Database path for failing-example storage\n",
+        "  --database <PATH|disabled|none>      Database path for failing-example storage\n",
```
DRMacIver: What does none do here that is distinct from disabled?

hegel-macros/src/lib.rs
```diff
@@ -18,6 +18,16 @@ use syn::{Data, DeriveInput, ItemFn, ItemImpl, parse_macro_input};
 pub fn derive_generator(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
 
+    if !input.generics.params.is_empty() {
+        return syn::Error::new_spanned(
+            &input.generics,
+            "#[derive(DefaultGenerator)] does not support generic types: the generated \
```
DRMacIver: This doesn't seem right. Shouldn't it support generic types but add a trait bound that the type parameter implements DefaultGenerator?

hegel-macros/src/stateful.rs
```diff
@@ -21,7 +21,15 @@ fn method_entries(methods: &[MethodInfo]) -> Vec<TokenStream> {
...
             let name = &m.name;
-            let attrs = &m.attrs;
+            // Only conditional-compilation attributes make sense on the
+            // generated vec entry; forwarding everything would put e.g. the
+            // method's doc comments on an expression, which trips
+            // unused_doc_comments in the user's crate.
+            let attrs: Vec<&Attribute> = m
+                .attrs
+                .iter()
+                .filter(|a| a.path().is_ident("cfg") || a.path().is_ident("cfg_attr"))
```
DRMacIver: I think this is the wrong behaviour and is far too restrictive. What about user-defined attribute style macros?

hegel-c/include/hegel.h
```diff
@@ -525,6 +532,9 @@ typedef struct hegel_run_result_t hegel_run_result_t;
  `hegel_settings_*` family of setters, hand to `hegel_run_start`, then
  free with `hegel_settings_free`. Settings can be reused across
  multiple runs; the engine reads them at `hegel_run_start` time.
+
+ A settings handle must only be used from one thread at a time (unlike
+ test-case handles, which detect and reject concurrent use).
```
DRMacIver: I think this isn't the intended semantics. settings should be usable from multiple threads as long as you don't modify them concurrently. A common use case is to create a settings object once and then reuse it, which may involve using it from multiple threads.

hegel-c/include/hegel.h
```diff
@@ -1031,10 +1053,10 @@ hegel_result_t hegel_pool_add(hegel_context_t *ctx,
  draws.
 
  On success writes the chosen variable id into `*out_variable_id` and
- returns `HEGEL_OK`. Returns `HEGEL_E_STOP_TEST` if the pool currently
+ returns `HEGEL_OK`. Returns `HEGEL_E_ASSUME` if the pool currently
  has no active variables — the caller should guard against that (e.g.
  only draw when it knows it has added at least one variable) or treat
- it like any other budget-exhaustion outcome.
+ it like a failed assumption and mark the test case INVALID.
```
DRMacIver: A failed assumption doesn't necessarily mark the test case as INVALID and the caller is allowed to resume from one (e.g. we do this in stateful testing)

PR #374, opened 2026-07-15
Move to polling for the stateful test lifecycle

rdck:
This shouldn't change the distribution of rule sequences much -- just moves the control into the engine. Internally, we still choose a step cap in the same way.

The cap is chosen lazily when the first rule is requested, following what we already do for swarm testing features. Alternatively, we could choose the step cap up front during state machine initialization. I assume this doesn't matter much, since we'll probably move away from an explicit step cap soon anyway?


src/stateful.rs
```diff
@@ -253,34 +246,20 @@ pub fn run<M: StateMachine>(mut m: M, tc: TestCase) {
...
-    let mut step = 0;
+    let mut steps_attempted: i64 = 0;
 
-    while steps_run_successfully < step_cap
-        && (is_single
-            || steps_attempted < 10 * step_cap
-            || (steps_run_successfully == 0 && steps_attempted < 1000))
-    {
-        step += 1;
+    while is_single || steps_attempted < 1000 {
```
rdck: Sanity check on this loop condition, please!

DRMacIver (approved):

PR #378, opened 2026-07-22
Add concurrent stateful testing

rdck:
This is still a draft; I want to do at least one more review pass before we merge.


src/run_lifecycle.rs
```diff
@@ -230,29 +258,39 @@ pub fn panic_message(payload: &Box<dyn std::any::Any + Send>) -> String {
 /// Also returns the caught panic payload for an `Interesting` result, so a
 /// final replay's caller can re-raise the test's *own* panic as the run's
 /// closing unwind instead of synthesizing one.
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn run_test_case(
```
Liam-DeVoe, line 273: let's blanket disable this in our clippy config. I dislike linters that warn about complexity or maintainability problems.

hegel-c/include/hegel.h
```diff
@@ -781,6 +788,28 @@ hegel_result_t hegel_settings_set_report_multiple_failures(hegel_context_t *ctx,
...
 
+/*
+ Declare the run nondeterministic: the test may produce different
+ outcomes (or draw different choice sequences) when run on identical
+ data — e.g. because it exercises real concurrency. The frontend must
+ set this whenever a run may be nondeterministic, typically because the
+ test uses concurrent stateful testing.
+
+ When set, the engine reports failures faithfully without attempting
+ anything that assumes deterministic replay: it skips data-tree
```
lmb: Nit: probably a claudism. Don't need to enumerate what is disabled.

hegel-c/include/hegel.h
```diff
@@ -781,6 +788,28 @@ hegel_result_t hegel_settings_set_report_multiple_failures(hegel_context_t *ctx,
                                                            hegel_settings_t *s,
                                                            bool yes);
 
+/*
+ Declare the run nondeterministic: the test may produce different
```
lmb: Isn't the core better placed to determine this?

hegel-c/include/hegel.h
```diff
@@ -1121,46 +1150,100 @@ hegel_result_t hegel_pool_generate(hegel_context_t *ctx,
...
- non-UTF-8 names.
+ `hegel_state_machine_next_group` / `hegel_state_machine_next_rule`
+ calls on the *same* test-case family. Returns `HEGEL_E_INVALID_ARG` if
+ `num_rules` or `num_groups` is zero, an entry of `rule_groups` is
+ outside `[0, num_groups)`, a group ends up with no rules,
+ `concurrency < 1`, or on null / non-UTF-8 names.
  */
 hegel_result_t hegel_new_state_machine(hegel_context_t *ctx,
                                        hegel_test_case_t *tc,
+                                       const char *const *group_names,
```
lmb: Is this optional? Why do groups need a name?

hegel-c/include/hegel.h
```diff
@@ -1121,46 +1150,100 @@ hegel_result_t hegel_pool_generate(hegel_context_t *ctx,
...
- with `HEGEL_STATUS_OVERRUN`).
+ Start the machine's next round: draw whether another round should run
+ at all and, if so, which concurrency group is current for it. Writes
+ the current group's index in `[0, num_groups)` into
+ `*out_group_index` when a new round has begun and the worker threads
+ should pull rules again — the index identifies the round's group, e.g.
+ for trace output — or `HEGEL_STATE_MACHINE_DONE` (-1) to indicate
+ termination of the whole state machine.
+
+ Call this on the *root* test-case handle at every join point — after
```
lmb: What is the root test case handle? The one we called hegel_new_state_machine on, presumably? Should be spelled out in docs.

hegel-c/include/hegel.h
```diff
@@ -1121,46 +1150,100 @@ hegel_result_t hegel_pool_generate(hegel_context_t *ctx,
...
+/*
+ Draw the index of the next rule for worker thread `thread_index` to run
+ this round, letting the engine choose the rule sequence. The returned
+ index is always a rule belonging to the current concurrency group (see
+ `hegel_state_machine_next_group`). Swarm testing is applied per thread:
+ a random subset of rules is enabled (at least one per group) on the
+ thread's first selection and selection is restricted to that subset for
+ the rest of the test case.
+
+ `thread_index` identifies the calling worker and must satisfy
```
lmb: We intermingle thread, group worker here. Would be best to stick to a consistent naming scheme.

hegel-c/include/hegel.h
```diff
@@ -1121,46 +1150,100 @@ hegel_result_t hegel_pool_generate(hegel_context_t *ctx,
...
+ `0 <= thread_index < concurrency` (passed at state-machine creation); a
+ thread index rather than the handle identifies the thread because a
```
lmb: This is confusing, because it doesn't tell me what testcase to pass in the first case.

I'm assuming that it's okay to pass the "root" testcase here, but that each worker needs a separate clone of the testcase?

Should testcases be long lived (so cloned once) or created and destroyed for each group?

hegel-c/include/hegel.h
```diff
@@ -1189,6 +1272,25 @@ hegel_result_t hegel_generate_boolean(hegel_context_t *ctx,
...
+ this is a dedicated primitive instead of a plain integer draw.
+
+ On success writes the drawn level into `*out_value` and returns
+ `HEGEL_OK`. Returns `HEGEL_E_STOP_TEST` when the engine's choice
+ budget is exhausted for this test case (the caller should abort the
+ body and call `hegel_mark_complete` with `HEGEL_STATUS_OVERRUN`).
+ Returns `HEGEL_E_INVALID_ARG` for a NULL `out_value` or
+ `max_value < 1`; the diagnostic is in `hegel_context_last_error`.
+ */
+hegel_result_t hegel_generate_concurrency(hegel_context_t *ctx,
```
lmb: Does this pay for itself? Seems easier to me to pass max_concurrency to new_state_machine?

hegel-c/include/hegel.h
rdck: I don't think so. It's the job of the frontend to do the scheduling, and the scheduling is where the nondeterminism comes from. If we say it's the job of the engine to do nondeterminism detection, then I don't see how we reasonably distinguish between unintended flakiness and intended nondeterminism.

hegel-c/include/hegel.h
rdck: That's definitely an option. I was thinking that it will be nice for the frontend to exert more control over the concurrency if it wants to; say, if the user is interested in a bug they know would be triggered at concurrency two but not at concurrency one. Maybe we pass both a min and max concurrency to new_state_machine?

hegel-c/include/hegel.h
rdck: Just for representation to the user. I was trying to be future proof for when we move more of that logic into the engine, but maybe we should just drop it for now; can always add it later.

hegel-c/include/hegel.h
lmb: Whenever we clone a testcase for example?

hegel-c/include/hegel.h
rdck: Done

src/run_lifecycle.rs
rdck: Good idea

hegel-c/include/hegel.h
rdck: I went ahead and removed it for now.

hegel-c/include/hegel.h
rdck: Yeah. Should be clear in the docs now.

hegel-c/include/hegel.h
rdck: Yep, done.

hegel-c/include/hegel.h
rdck: Sorry, what do you mean by the first case here?

Yes, each worker needs its own clone, and clones should be long-lived. This should be clearer now.

hegel-c/include/hegel.h
rdck: Let's discuss in person.

hegel-c/include/hegel.h
rdck: I went ahead and removed this in favor of min/max concurrency bounds at machine initialization time.

rdck:
Notably absent is support for singleton rules (rules that can't overlap any other rules, including themselves). Open to suggestions, because I haven't been able to come up with an interface I like that supports that case.

hegel-c/include/hegel.h
```diff
@@ -1121,46 +1139,128 @@ hegel_result_t hegel_pool_generate(hegel_context_t *ctx,
...
- `hegel_state_machine_next_rule` which rule to run at each step and
- applies it, until that call signals that no more steps should
- follow.
-
- On success writes the new machine's id into `*out_state_machine_id`
- and returns `HEGEL_OK`. The id is opaque; pass it to subsequent
- `hegel_state_machine_next_rule` calls on the *same* test case.
- Returns `HEGEL_E_INVALID_ARG` if `num_rules` is zero, or on null /
- non-UTF-8 names.
+ testing, sequential or concurrent: `num_groups` concurrency groups
```
lmb: Somwhat of a nit: in theory we don't need to predeclare group numer, it could just be `max(rule_groups)`. Might be a bit easier to understand as well.

hegel-c/include/hegel.h
```diff
@@ -1121,46 +1139,128 @@ hegel_result_t hegel_pool_generate(hegel_context_t *ctx,
...
+ Creating the machine draws from the calling handle's stream: the
+ concurrency level, the test case's round cap, and each worker's swarm
+ parameters are decided here, up front, so the machine is fully
+ constructed before any rule is requested.
+
+ On success writes the new machine's id into `*out_state_machine_id`,
+ the drawn concurrency level into `*out_concurrency`, and returns
+ `HEGEL_OK`. The id is opaque; pass it to subsequent
+ `hegel_state_machine_next_group` / `hegel_state_machine_next_rule`
+ calls on the *same* test-case family. Returns `HEGEL_E_STOP_TEST` when
```
lmb: What happens if we use the wrong test-case family? Is this a misconfiguration we can diagnose with an error? If this silently causes non-shrinkability or similar it points in the direction that the API we have is prone to misuse.

hegel-c/include/hegel.h
lmb: Also, from a didactic point: I would lead this with a high level description of the execution model instead of diving into the arguments straight away.

lmb (commented):
There's probably stuff we can improve in the C API, but I'd say ship it!

hegel-c/include/hegel.h
rdck: Yes, it's probably better if the group identifiers are just unique integers and not required to be contiguous, too. I'll fix that.

rdck: Changed in 0c741dedea09485d87693ae7cce45f1531884734.

lmb (approved):

PR #387, opened 2026-08-03
Keep libhegel resident with -z nodelete to fix a dlclose crash

DRMacIver:
> _Draft — over to you for the title and a short description._

<details>

### What

Adds `hegel-c/build.rs`, which builds the `libhegel` cdylib with `-Wl,-z,nodelete` on ELF platforms (Linux/Android). The flag applies only to the cdylib artifact; the `staticlib`/`rlib` are untouched.

### Why

An embedder that `dlopen`s libhegel, runs the engine, and then `dlclose`s it while a thread that called into it is still alive could `SIGSEGV` during thread teardown. It was most reliably triggered by a run that persists to the on-disk failure database.

Root cause (from a coredump backtrace):

- `DirectoryTestCaseDatabase::save` writes atomically via `tempfile::NamedTempFile`, which uses `fastrand`'s thread-local RNG.
- Initialising that thread-local runs `std::thread::current::init_current`, which registers std's **thread-local-destructor pthread key** on the *caller's* thread — with the destructor code living inside libhegel.
- glibc records that destructor in the process-global `__pthread_keys` and calls it when the thread exits.
- If the embedder has `dlclose`d libhegel by then, the code is unmapped and the destructor pointer dangles → segfault during thread teardown. (The crashing PC matched `std::sys::thread_local::guard::key::enable::run` in the unmapped range exactly.)

The same hazard applies to the global panic hook the engine installs. `-z nodelete` turns `dlclose` into a no-op for this library, so it stays mapped and all such destructors remain valid for the life of the process — the standard fix for a `dlopen`-able Rust cdylib. On macOS/Windows a `dlopen`ed dynamic library isn't eagerly unmapped the same way, so no equivalent is needed there.

### Regression coverage

The existing `libhegel_replays_persisted_failure_with_same_database_key` smoke test exercises this exact path (persist on run 1, replay on run 2, then drop the `Library`). It segfaulted without the flag and passes with it; a doc comment on the test now records the dependency.

Verified by control: rebuilt the cdylib with the flag removed → the smoke test `SIGSEGV`s again; restored → passes. `FLAGS_1: NOW NODELETE` confirmed via `readelf -d`. Full `hegel-c` suite green (1120 unit + 16 smoke).

### Notes

- This is a pre-existing crash, not introduced by any recent change — it reproduces on `main` (verified on `ecff302d`). It appears to surface on aarch64/glibc; other targets may tear TLS down in an order that hides it.
- `RELEASE_TYPE: patch` (`hegel-c/RELEASE.md`).
</details>

### DIFF
--- a/hegel-c/RELEASE.md
+++ b/hegel-c/RELEASE.md
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch fixes a crash (SIGSEGV) that could occur when an application `dlopen`s libhegel, runs a test, and then `dlclose`s the library while a thread that called into it is still alive. It was most reliably triggered by a run that persists to the on-disk failure database.
+
+Running the engine registers `std` thread-local destructors — and a global panic hook — whose code lives inside libhegel. glibc runs those thread-local destructors when the thread exits; if the library has already been unloaded by then, the destructor pointers dangle and the process crashes during teardown. libhegel is now built with `-z nodelete` on ELF platforms (Linux/Android), so `dlclose` keeps it resident and those destructors stay valid for the life of the process.
--- a/hegel-c/build.rs
+++ b/hegel-c/build.rs
@@ -1,0 +1,26 @@
+fn main() {
+    // Mark the cdylib as non-deletable so `dlclose` keeps it resident for the
+    // rest of the process's life.
+    //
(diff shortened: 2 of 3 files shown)


DRMacIver:
(I'm going to fix this properly instead)

PR #398, opened 2026-08-06
Inform the engine of failed rules

rdck:
Correct accounting of failed rules got lost in https://github.com/hegeldev/hegel-rust/pull/374. This fixes that.


DRMacIver (approved):
LGTM, though I snuck in a large change before you (sorry) that you'll need to rebase/merge onto and it might surface some issues.

PR #401, opened 2026-08-09
#[composite] macro improvements

brownben:
This PR fixes some minor papercuts introduced by the changes to the #[composite] macro in #397.

---

    #[composite] Better support for borrowed parameters

    Previously, functions using the #[composite] macro and references worked
    similarly to regular functions, and lifetimes could be elided in the same
    places.

    With the changes to store the parameters in a struct, if we pass a parameter
    by reference in we have to name the lifetime even if it is not used in the
    output. This makes functions with the composite macro act slightly differently
    to "normal" functions.

    If any parameters are passed as a reference without a lifetime, the macro
    now rewrites to add a hidden lifetime so it can be stored without the user having
    to specify a lifetime. Return types are left unchanged, and if they have elided
    lifetimes, the user will get the same warning as regular functions that the compiler
    can't tell if the lifetime should be from the test case or their parameter.

---

    #[composite] Stop items_after_statements warning

    If you define a constant, or function in a function using the
    `#[hegel::composite]` macro, you currently get a warning from
    the clippy `items_after_statements` lint.

    This is due to hegel injecting `::hegel::__assert_is_test_case::<::hegel::TestCase>();`
    as the first item in the function.

    Update to insert the whole body block after a block containing the
    injected item, rather than inserting the function in the user's
    block.
    
---

    #[composite] Output a generator even if there is an error

    Previously, if an error was found during the composite macro
    expansion (e.g. not having the test case as reference), only
    an error would have been returned.

    If the generator was used anywhere, it would spread many
    compiler warnings up the chain.

    Update so the macro tries to return the generator (and the errors)
    if an error is found, so anything which depends on it does not
    bring up an error.

    This also allows the macro to return multiple errors if necessary
    (e.g. both missing return type and wrong test case argument).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,3 +4,4 @@ This patch brings various minor improvements to the `#[composite]` macro.
 
 - Adds support for passing parameters by reference without an explicit lifetime, if the lifetime is not used in the return type.
 - Stops the items_after_statements clippy lint from firing if constants/ functions are defined in the function body.
+- Tries to output generator even if errors are found as to not raise errors elsewhere
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch brings various minor improvements to the `#[composite]` macro.
+
+- Adds support for passing parameters by reference without an explicit lifetime, if the lifetime is not used in the return type.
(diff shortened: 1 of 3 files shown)


DRMacIver:
Thanks! Much appreciated, and sorry about the papercuts.

Issue #75, opened 2026-03-18
Consider adding `hegel::prelude`


Liam-DeVoe:
It might be nice to support `use hegel::prelude::*` to bring common imports into scope. Here's what I would put in it, if we did this:

```rust
pub use crate::TestCase;
pub use crate::generators::{self, Generator};
pub use crate::composite;
```

We should explicitly not import:
* `crate::test` (aliases standard `#[test]`)

PR #412, opened 2026-08-15
feat: add prelude for common imports

sweb:
Hi,

addressing #75 - this comes with 2 commits. The first commit basically does what the issue asks for. The second commit also adds the `gs` alias to the prelude - I am not sure here, as `gs` is not very self descriptive but all examples use it and without it the added value of a prelude seems smaller than it could be as it would in most cases require a second import. 

Happy to drop the second commit. I can also take `Not relevant anymore, pls close :)`

just targets are all green.

Disclaimer: Claude helped me here a bit - I would have missed the changelog file.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,7 +4,6 @@ This patch adds `hegel::prelude`, which brings the items most tests need into sc
 
 ```rust
 use hegel::prelude::*;
-use hegel::generators as gs;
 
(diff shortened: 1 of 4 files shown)


sweb:
@Liam-DeVoe may I ask you for a review? Also completely fine to say that this is no longer needed :)

PR #430, opened 2026-09-01
Extend one_of! to arbitrary numbers of arms

DRMacIver:
It turns out proptest users often have really large branches in their one_of. Rather than continuing extending the number of alternatives up to 30ish (I did a survey. There are some 200+ ones but those are all sparkling sampled_from, but there are real 30) this extends it out to infinity by using a typed-level cons-list. 

<details>
<summary>PR description (AI generated)</summary>

`one_of!` now covers arities 1–30; real-world grammars (e.g. a 24-arm expression table) exceeded the old limit of 12. Components stay unboxed, and the result is printable exactly when every component is.

For anything larger, the fixed-arity machinery is exported as `impl_one_of!`, and the `generators::draw_one_of` helper its expansions share is now public. The compile error past 30 arms names both escape hatches.

</details>


src/generators/combinators.rs
```diff
@@ -67,7 +67,10 @@ pub struct OneOfGenerator<'a, T, B = BoxedPrintableGenerator<'a, T>> {
 /// alternative dispatch (and whether it draws silently or printing)
 /// injected. Using this from both draw paths is what keeps their choice
 /// streams identical.
-fn draw_one_of<T>(tc: &TestCase, max_index: usize, draw_at: impl FnOnce(usize) -> T) -> T {
+///
+/// Public so that generators defined with [`impl_one_of!`](crate::impl_one_of)
+/// (and any hand-written `one_of`-shaped generator) share it.
+pub fn draw_one_of<T>(tc: &TestCase, max_index: usize, draw_at: impl FnOnce(usize) -> T) -> T {
```
DRMacIver: Should be doc(hidden) though.

PR #450, opened 2026-09-04
Invert stop condition for stateful tests

### LINKED ISSUE #433: Stateful test with high step count and multiple invariants now triggers `LargeInitialtestCase`
Stateful tests which previously passed now fail with "FailedHealthCheck: LargeInitialTestCase".

It seems to be the combination of a high `stateful_step_count` and many invariants. If I reduce either, the tests pass (and reducing the number of rules has no effect).
 
Even examples which don't draw much data in the test now fail with the health check.

I think it must be related to the recent invariant changes, as it works in 0.34.1 but fails in 0.35.0.

## Minimal Example

```
use hegel::TestCase;
use hegel::generators as gs;

struct IntegerStack {
    stack: Vec<i32>,
}

#[hegel::state_machine]
impl IntegerStack {
    #[rule]
    fn push(&mut self, tc: TestCase) {
        let integers = gs::integers::<i32>;
        let element = tc.draw(integers());
        self.stack.push(element);
    }

    #[rule]
    fn pop(&mut self, _: TestCase) {
        self.stack.pop();
    }

    #[invariant]
    fn invariant_a(&self, _: TestCase) {}

    #[invariant]
    fn invariant_b(&self, _: TestCase) {}

    #[invariant]
    fn invariant_c(&self, _: TestCase) {}

    #[invariant]
    fn invariant_d(&self, _: TestCase) {}

    #[invariant]
    fn invariant_e(&self, _: TestCase) {}
}

#[hegel::test(stateful_step_count = 500)]
fn test_integer_stack(tc: TestCase) {
    let stack = IntegerStack { stack: Vec::new() };
    hegel::stateful::run(stack, tc);
}
```

```
running 1 test
test test_integer_stack ... FAILED

failures:

---- test_integer_stack stdout ----

thread 'test_integer_stack' (57244) panicked at /home/ben/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hegeltest-0.35.0/src/run_lifecycle.rs:587:13:
FailedHealthCheck: LargeInitialTestCase — the smallest natural input for this test is very large, which makes it hard to generate and shrink good inputs. Consider reducing the amount of data generated, or introducing small alternatives (e.g. `gs::one_of` with an empty option). If this is expected, suppress the check with suppress_health_check = [HealthCheck::LargeInitialTestCase].
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

rdck:
I had noticed this before but thought we had fixed it already. Fixes https://github.com/hegeldev/hegel-rust/issues/433.


rdck:
(The reason https://github.com/hegeldev/hegel-rust/issues/433 started failing in 0.35.0 is that `check_invariants` started consuming entropy.)

PR #454, opened 2026-09-05
Suppress TooSlow by default in CI

valwalt:
Hegel already detects CI to derandomize runs and disable the example database. This adds `TooSlow` to the same defaults in both `hegeltest` and `hegeltest-c`.

Hypothesis has suppressed [`HealthCheck.too_slow` in its automatically selected CI profile](https://github.com/HypothesisWorks/hypothesis/commit/13c3785854da056387aeac789300537e501a3c14) since 2024. Explicit `suppress_health_check` settings still replace the CI default, including an empty list that re-enables the check.


DRMacIver:
Thanks! Pure oversight on our part that we weren't already doing this, appreciate the catch.

PR #462, opened 2026-09-07
Add a reproduction pointer to failure output

### LINKED ISSUE #438: Failure output contains no reproduction pointer
A failing test prints the shrunk draws and the panic, then stops. No seed, no blob, no mention that a `.hegel/` database was written or that a rerun will replay the failure. All six porting agents flagged this; the Hypothesis "You can reproduce this example by..." line is the feature they missed most.

The machinery exists but nothing routes to it: `Settings::print_blob` defaults to `false` (`src/runner.rs:163`) and has no environment override — the only recognised variables are `HEGEL_DATABASE`, `HEGEL_STATISTICS`, and `HEGEL_TEST_CASES` — and `#[hegel::reproduce_failure]` is only discoverable from the crate docs. For "failed in CI, reproduce locally" there is no actionable output at all unless the test author preconfigured `print_blob`.

**Repro:** make any hegel test fail and read the output.

**Suggested shape:** one trailer line on failure saying where the example was saved (or that saving was disabled), and/or print the blob when no database is available.

---
*Found during the pretty-printing ergonomics work (`DRMacIver/pretty-printing-ergonomics` @ d7e2a765), by agents re-porting the hegel-rust-oss-bug-finding suites from hegeltest 0.28.2. Issue text drafted by Claude.*

DRMacIver:
Fixes #438.

<details>
<summary>Description written by Claude (AI generated)</summary>

A failing test previously printed the shrunk draws and the panic, then stopped. There was no pointer to the saved example, the reproduce blob, or `#[hegel::reproduce_failure]`.

Failure output now ends with a reproduction pointer:

- When the failing example was saved, one line names the database directory and says a rerun replays it:

  ```
  The failing example was saved to '.hegel/examples'. Rerunning this test will replay it.
  ```

- When nothing was saved (database disabled, as in CI by default, or no database key), the reproducer line that used to require `print_blob` is printed instead. Macro-defined tests get the `#[hegel::reproduce_failure("…")]` attribute wording, and bare `hegel()` / `Hegel::new` runs get the builder wording:

  ```
  To reproduce this failure, call reproduce_failure on the test's builder:
      Hegel::new(...).reproduce_failure("AAEAAAAACgEAAAAA")
  ```

`print_blob = true` still forces the reproducer line, including on quiet runs. Quiet runs print neither line otherwise.

The engine owns the database, so it reports the persisted directory on the run result (`TestRunResult::database_path`, exposed as the new `hegel_run_result_database_path` C ABI function), and the frontend renders the trailer in `run_lifecycle::drive`.
</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+This patch adds a reproduction pointer to failure output
+([#438](https://github.com/hegeldev/hegel-rust/issues/438)). A failing test
+now ends with a line naming the database directory the shrunk example was
+saved to. When nothing was saved (the database is disabled, as in CI by
+default, or the test has no database key) the copy-pasteable
+`reproduce_failure` line is printed instead. Previously that line required
+enabling the `print_blob` setting.
--- a/hegel-c/RELEASE.md
+++ b/hegel-c/RELEASE.md
@@ -1,0 +1,5 @@
+RELEASE_TYPE: patch
+
+This patch adds `hegel_run_result_database_path`, which reports the database
+directory a failing run saved its counterexamples to (NULL when nothing was
+saved), so frontends can point at the saved example in their failure output.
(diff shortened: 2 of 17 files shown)


DRMacIver:
I think this is a mistake. There are probably some ergonomics we could improve here but this isn't it.

PR #466, opened 2026-09-07
Anchor the default failure database at the crate root

### LINKED ISSUE #439: Default failure database path resolves against the runtime cwd; chdir tests lose failures
The default database is the relative path `.hegel/examples` (`hegel-c/src/native/test_runner.rs:992`), resolved at failure time against the process's current directory. Tests that chdir lose failures permanently.

Observed in the glob port: `tests/glob-std.rs` has a pre-existing test that does `env::set_current_dir(tempdir)` and runs concurrently in the same binary. When a property test in that binary failed, `.hegel/` was written inside the tempdir and destroyed with it — no database anywhere afterwards, and the next run passed. Running the property test filtered (so the chdir test never ran) wrote `.hegel/` in the package root and replay worked.

**Repro:** in one test binary, have test A `set_current_dir` into a `TempDir` it drops, and hegel test B fail while A holds the cwd.

**Suggested fix:** anchor the default at `CARGO_MANIFEST_DIR`.

---
*Found during the pretty-printing ergonomics work (`DRMacIver/pretty-printing-ergonomics` @ d7e2a765), by agents re-porting the hegel-rust-oss-bug-finding suites from hegeltest 0.28.2. Issue text drafted by Claude.*

DRMacIver:
Fixes #439.

<details>
<summary>Description (written by Claude)</summary>

The default failure database was the relative path `.hegel/examples`, resolved against the process cwd at failure time (`hegel-c/src/native/test_runner.rs`), so a concurrently running test that chdir'd into a tempdir made hegel write `.hegel/` into that tempdir, which was then destroyed with the failure in it.

`CARGO_MANIFEST_DIR` is Rust-specific, so the anchoring lives in the Rust frontend and the engine default is unchanged. The `#[hegel::test]`, `#[hegel::main]`, and standalone-function expansions now pass `env!("CARGO_MANIFEST_DIR")`, captured at the user crate's compile time, to the runner through a hidden `__database_root` builder method. Runs built directly with `Hegel::new` fall back to the runtime `CARGO_MANIFEST_DIR`, and the engine's cwd-relative default applies only when neither is available. Only the unset default is anchored: paths configured via `Settings::database`, `HEGEL_DATABASE`, or `--database` behave as before, including relative ones.

One consequence for this repo's own suite: failing tests that use the default database now write to the repo root `.hegel/` (gitignored) when run locally, instead of the per-binary tempdir cwd. The suite's helpers disable the database wherever a test asserts on generation behaviour, so this only affects tests that merely require a failure, and a replayed failure satisfies those.

Tests: an end-to-end regression test in `tests/test_settings.rs` that fails before this change, unit tests for the anchoring and root-resolution logic in `tests/embedded/runner_tests.rs`, and expansion tests for the three macros under `hegel-macros/tests/embedded/`. The pre-existing engine-default test in `tests/embedded/runner_tests.rs` now removes `CARGO_MANIFEST_DIR` from its child's environment, since it exercises the cwd fallback.
</details>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+This patch fixes where the default failure database is created ([#439](https://github.com/hegeldev/hegel-rust/issues/439)). It was the relative path `.hegel/examples`, resolved against the process's current directory at failure time, so a test that changed the current directory (or ran concurrently with one that did) could write `.hegel/` into a temporary directory that was then deleted, losing the stored failure. The default is now anchored at the crate root, taken from `CARGO_MANIFEST_DIR` at the test's compile time or from the environment at runtime. Explicitly configured database paths are unchanged.
--- a/hegel-macros/src/hegel_main.rs
+++ b/hegel-macros/src/hegel_main.rs
@@ -112,6 +112,7 @@ pub fn expand_main(attr: TokenStream, item: TokenStream) -> TokenStream {
             .settings(__hegel_settings)
             .__single_test_case()
             .__database_key(format!("{}::{}", module_path!(), #fn_name))
+            .__database_root(env!("CARGO_MANIFEST_DIR").to_string())
             .test_location(::hegel::TestLocation {
                 function: #fn_name.to_string(),
                 file: file!().replace('\\', "/"),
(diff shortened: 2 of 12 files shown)


DRMacIver:
I think this is a good idea but I'm going to roll it into some other work I'm doing.

PR #468, opened 2026-09-07
Fix papercuts when using Hegel for long-running workloads

rdck:
At the very least we should turn off the `TooSlow` health check for workloads. I suspect that's not the only thing needed, though.


rdck:
For now, this

+ suppresses both `TooSlow` and `TestCasesTooLarge` for main functions
+ lowers the state machine stop probability

We'll separately need to change or remove the choice limit, to support long-running test cases.

PR #478, opened 2026-09-09
Add minimal WebAssembly support to hegel-c

Xapollon430:
## Summary

- add `wasm32-unknown-unknown` support for `hegel-c`;
- preserve the existing native C ABI and native platform behavior;
- provide Wasm host imports for entropy and monotonic time;
- disable the filesystem-backed failure database and concurrent state machines in Wasm;
- add Wasm-only allocation helpers for browser TypeScript pointer marshalling;
- compile the Wasm test suite in CI and smoke-test the host boundary;
- publish `libhegel-wasm32-unknown-unknown.wasm` and its checksum with hegel-c releases;
- add a `hegel-c/RELEASE.md` entry and `just c-build-wasm` documentation.

Wasm uses the default PRNG backend seeded through `entropy_fill`; the direct urandom backend is unavailable and follows the existing fallback behavior. Antithesis, filesystem persistence, and Wasm threads are not enabled in this phase.

## Validation

- `cargo test -p hegeltest-c --lib` — 1,319 passed
- `cargo check -p hegeltest-c`
- `cargo check -p hegeltest-c --target wasm32-unknown-unknown`
- `cargo test -p hegeltest-c --target wasm32-unknown-unknown --lib --no-run`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`
- release Wasm build and temporary Node host-boundary smoke check

No standalone Wasm smoke-test files are included; the smoke check runs inline in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Install Rust and WebAssembly target
         run: |
-          rustup toolchain install stable --profile minimal
+          rustup toolchain install stable --profile minimal -c llvm-tools-preview
           rustup default stable
           rustup target add wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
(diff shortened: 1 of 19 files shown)


DRMacIver:
On a quick skim (will do a more thorough review later) basically LGTM, but one big thing missing from this is CI. We should be running our test suite on wasm and shipping wasm builds as part of the releases.

DRMacIver:
Oh yes also it needs a release file.

Xapollon430:
> On a quick skim (will do a more thorough review later) basically LGTM, but one big thing missing from this is CI. We should be running our test suite on wasm and shipping wasm builds as part of the releases.

I added it to CI and also added a release file. Let me know if those look good, thanks for the quick turn around

valwalt:
This is very cool! I have already experimented with hacking in WASM for my Swift wrapper with some success, so I gave this PR a try. Two questions:

1. The static archive currently refers to Rust-mangled names for the entropy and clock hooks, so I can’t reliably supply them from Swift. Could those hooks be given stable C symbol names for static builds?
2. And relatedly, could we publish the `.a` alongside the `.wasm`? I believe Rust already generates this, which would allow me to link it with SwiftPM.

DRMacIver:
These seem like reasonable things to add to me. @Xapollon430 Can you do them as part of this PR, or would you like us to do a followup one?

Xapollon430:
I implemented the stable naming for the monotonic and entropy hook. Also added the `.a` distributable and added it to CI. Let me know if it looks good to you guys.

PR #481, opened 2026-09-09
Pass the stateful step count to the run call instead of settings

rdck:
See release note. @DRMacIver any opinion on the naming?


rdck:
We decided offline to go with a builder interface for configuring stateful tests.

PR #487, opened 2026-09-10
Rework choice limit

rdck:
See release note.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,21 +1,12 @@
-RELEASE_TYPE: minor
+RELEASE_TYPE: patch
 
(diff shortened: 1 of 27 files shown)


DRMacIver:
I don't necessarily object to this, but the reason why we deprecated the equivalent setting in Hypothesis is that this number is quite hard for users to reason about, and I think that's true here too - it's not really obvious what a particular number means, or what you should set it to to get a particular outcome.

Was there a specific reason you went with this approach instead of the health check one we talked about yesterday?

(The ID change seems strictly good though)

rdck:
No, that was just an artifact of an earlier draft. I do think we should probably expose it from the engine, though. I think most frontends will just want a toggle, but we may as well make it configurable if frontend authors want control.

rdck:
We decided offline that the engine should just expose a toggle, rather than a configurable choice budget.

PR #489, opened 2026-09-11
Renamed float_clamp to float_restrict_and_redraw as float_clamp wasn't clamping in the traditional sense

max-amb:
Very open to an alternative name.


DRMacIver:
@max-amb PR looks good to me but you need to add a release file.

max-amb:
Ah I thought because it was internal it wasn't needed. Doing now.

max-amb:
Oops... git is hard.

PR #500, opened 2026-09-20
Failing test to show internal errors cause test failure but with no useful output

keirwl:
Just a quick failing case to demonstrate that internal errors aren't re-raised as per #47


DRMacIver:
Thanks for the report! I've rolled your test into #502, which has a fix for this.

